PR #672, opened 2026-04-20
fix(ci): SMI-4361 publish.yml extracts nested per-package node_modules

wrsmith108:
## Summary
- `publish.yml` Validate job has been failing since SMI-4293 Wave 5C introduced `web-tree-sitter` — extract-node_modules step silently failed for every package because `docker cp` destination parent `/tmp/packages/<pkg>/` was never created.
- Align publish.yml's extract step with the already-working pattern in `ci.yml`/`e2e-tests.yml` (SMI-2159): `mkdir -p /tmp/packages/$pkg` + trailing-slash cp + visibility logging + cleanup guard.
- Make `workflow_dispatch -f dry_run=true` actually exercise Validate (previously a no-op when all versions already published).

## Root cause
See `docs/internal/implementation/smi-4361-publish-yml-web-tree-sitter.md` (pending submodule commit). In short: npm nested `web-tree-sitter@0.25.10` at `packages/core/node_modules/web-tree-sitter` with no root-hoisted copy (core-only dep). `publish.yml` extract step failed to mkdir the per-package parent before `docker cp`, so nested node_modules never made it into the artifact. Validate job bind-mounted the empty workspace over `/app`, tsc TS2307'd.

Why it only broke now: zod and @opentelemetry/sdk-trace-base are also nested under core, but they have root-hoisted copies from other workspaces that TypeScript's upward module resolution silently falls back to. `web-tree-sitter` is the first core-only nested dep with no fallback.

Why the `@astrojs/vercel` case (SMI-4209 comment) wasn't caught: website build canary is `continue-on-error: true`.

## Verification
This PR is self-testing. Dispatched `publish.yml -f dry_run=true` against this branch at run [24647022347](https://github.com/smith-horn/skillsmith/actions/runs/24647022347). With the second commit's guard relaxation, validate now runs on dry_run=true — so a green run proves the extract step correctly pulls `packages/core/node_modules/web-tree-sitter` into the artifact and tsc resolves the module.

## Test plan
- [x] Static: bash syntax + YAML parse verified locally
- [x] Static: new pattern is byte-for-byte aligned with ci.yml:696-705 (battle-tested)
- [ ] CI: `publish.yml -f dry_run=true` completes Validate successfully (in progress)
- [ ] Publish jobs remain gated on `dry_run=false` or `release` event — no accidental publishes
- [ ] `[skip-impl-check]`

[skip-impl-check]

Closes SMI-4361.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wrsmith108:
## Verification result — dry-run [24647022347](https://github.com/smith-horn/skillsmith/actions/runs/24647022347)

| Step | Before this PR | After this PR |
|---|---|---|
| `Extract node_modules` | ✓ (silently dropped nested deps) | ✓ (correctly extracts all per-package `node_modules`) |
| `Build publishable packages (Docker)` | ❌ TS2307 on web-tree-sitter | ✅ **success** |
| `Website build canary` | n/a (non-gating) | ✅ success |
| `Run tests (Docker)` | ❌ SIGTERM at ~27s | ❌ SIGTERM at ~27s |

**The original SMI-4361 defect is fixed.** The `Build publishable packages (Docker)` step — which was failing with `TS2307: Cannot find module 'web-tree-sitter'` — is now green. The nested `packages/core/node_modules/web-tree-sitter` extraction is working correctly (confirmed by passing tsc against core, mcp-server, cli, enterprise).

The `Run tests (Docker)` failure is **pre-existing SMI-4244** (not a regression from this PR): vitest gets SIGTERM'd at ~27s inside the publish.yml Validate container. This has been the documented reason for using the local publish fallback since 2026-04-16. Fixing SMI-4244 is out of scope for this PR — if anything, this PR makes SMI-4244 *more visible* (previously the build step masked it).

### What this unblocks
- Future `web-tree-sitter`-class nested deps in any workspace will now extract correctly.
- `gh workflow run publish.yml -f dry_run=true` now actually exercises docker-build + Validate, making it a real smoke test instead of a no-op.

### What remains
- SMI-4244 (vitest SIGTERM in publish.yml) still blocks CI-driven publishes. Local fallback remains the path until that's separately resolved.

PR #1365, opened 2026-05-26
fix(docker): entrypoint npm rebuild blocked by .npmrc ignore-scripts=true (SMI-5200)

wrsmith108:
## Summary

- Root cause confirmed by Peter Lee (open source contributor, 2026-05-25): `ignore-scripts=true` in `.npmrc` is bind-mounted into the container at `/app/.npmrc`, causing `npm rebuild hnswlib-node` to skip node-gyp entirely and exit 0 with no binary produced
- `hnswlib-node@3.0.0` ships source-only (`gypfile: true`) and requires compilation; the other three native modules use prebuilt binaries and are unaffected
- Fix: add `--ignore-scripts=false` to override `.npmrc` in the rebuild loop; replace `2>/dev/null || true` with visible output + `set -e`-safe error handling

## Test plan

- [ ] `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` — workaround still valid pre-merge
- [ ] After merging: start fresh container with `.npmrc` `ignore-scripts=true` present on host; all four native modules should validate ✓
- [ ] Force a rebuild failure (e.g. temporarily remove g++ in container) — error now surfaces in logs instead of being swallowed
- [ ] `docker exec skillsmith-dev-1 npm run preflight` green

## Notes

Workaround for contributors until merged:
```
docker compose --profile dev down
docker volume rm skillsmith_node_modules
docker compose --profile dev up -d
```

[skip-impl-check]

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -267,6 +267,7 @@ Project skills load from the `.claude/skills/` mount-point of the `skillsmith-st
 |---------|-----|
 | Container won't start | `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` |
 | Native module errors | `docker exec skillsmith-dev-1 npm rebuild better-sqlite3 onnxruntime-node` |
+| `hnswlib-node` fails validation after rebuild (`ignore-scripts=true` in `.npmrc` blocks node-gyp) | `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` — removes stale volume so Docker re-initialises from image layer (compiled without `.npmrc`). Permanent fix: SMI-5200 |
 | Platform mismatch (SIGKILL 137) | `rm -rf packages/*/node_modules/better-sqlite3 packages/*/node_modules/onnxruntime-node` then rebuild |
 | Node ABI mismatch | WASM fallback auto-activates (core ≥0.4.10). Restore native: rebuild in Docker + `./scripts/repair-host-native-deps.sh` (SMI-4549) |
 | "invalid ELF header" in Docker (SMI-4698) | [git-crypt-guide.md § Host Native Bindings](.claude/development/git-crypt-guide.md#host-native-bindings--sessionstart-instrumentation-smi-4549) |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -267,7 +267,7 @@ Project skills load from the `.claude/skills/` mount-point of the `skillsmith-st
 |---------|-----|
 | Container won't start | `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` |
 | Native module errors | `docker exec skillsmith-dev-1 npm rebuild better-sqlite3 onnxruntime-node` |
-| `hnswlib-node` fails validation after rebuild (`ignore-scripts=true` in `.npmrc` blocks node-gyp) | `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` — removes stale volume so Docker re-initialises from image layer (compiled without `.npmrc`). Permanent fix: SMI-5200 |
+| `hnswlib-node` fails validation after rebuild (`ignore-scripts=true` in `.npmrc` blocks node-gyp — see SMI-5200) | Permanent fix ships in PR #1365. Until merged, delete the stale volume so Docker re-initialises `node_modules` from the image layer (built without `.npmrc`): `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` |
 | Platform mismatch (SIGKILL 137) | `rm -rf packages/*/node_modules/better-sqlite3 packages/*/node_modules/onnxruntime-node` then rebuild |
 | Node ABI mismatch | WASM fallback auto-activates (core ≥0.4.10). Restore native: rebuild in Docker + `./scripts/repair-host-native-deps.sh` (SMI-4549) |
 | "invalid ELF header" in Docker (SMI-4698) | [git-crypt-guide.md § Host Native Bindings](.claude/development/git-crypt-guide.md#host-native-bindings--sessionstart-instrumentation-smi-4549) |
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -149,6 +149,7 @@ if [ $VALIDATION_FAILED -eq 1 ]; then
     if [ $REBUILD_FAILED -eq 1 ]; then
         echo -e "${RED}[entrypoint] Native module validation failed after rebuild.${NC}"
         echo -e "${YELLOW}Try: docker compose down && docker compose build --no-cache${NC}"
+        echo -e "${YELLOW}For verbose rebuild output (run on host): docker exec <container> npm rebuild hnswlib-node${NC}"
```
wrsmith108: **Bug: hardcoded module name + non-functional placeholder**

Two issues on the same line:

1. `hnswlib-node` is hardcoded, but `REBUILD_FAILED=1` is set for **any** of the four modules at lines 143–145. If `better-sqlite3` or `esbuild` fails rebuild, this hint points the user at the wrong module.

2. `<container>` is a literal placeholder that will fail copy-paste. Every other reference in this file and CLAUDE.md uses the concrete name `skillsmith-dev-1`.

Suggested fix — make the hint dynamic and concrete:
```suggestion
        echo -e "${YELLOW}For verbose rebuild output (run on host): docker exec skillsmith-dev-1 npm rebuild ${FAILED_MODULES}${NC}"
```
where `FAILED_MODULES` is built up in the re-validation loop (lines 141–147) — e.g. append each failing module's name to a space-separated string and reference it here.

docker-entrypoint.sh
```diff
@@ -134,7 +134,7 @@ if [ $VALIDATION_FAILED -eq 1 ]; then
 
     for module in "${NATIVE_MODULES[@]}"; do
         echo -e "${YELLOW}  Rebuilding ${module}...${NC}"
-        npm rebuild "${module}" 2>/dev/null || true
+        npm rebuild "${module}" --ignore-scripts=false || echo -e "${YELLOW}  ↳ npm rebuild exited non-zero for ${module} (see output above)${NC}"
```
wrsmith108: **Stale comment at lines 112–116**

The block comment just above (lines 112–116) reads:

> *With .npmrc ignore-scripts=true (SMI-4672), this loop is the only runtime rebuild path on cache miss / ABI mismatch*

This now describes a resolved problem — line 137 adds `--ignore-scripts=false` precisely to fix this. A maintainer reading the comment will conclude the issue is still open.

Suggested comment update:
```
# With .npmrc ignore-scripts=true (SMI-4672), this loop is the only runtime rebuild path on cache miss / ABI
# mismatch. Each rebuild uses --ignore-scripts=false (SMI-5200) to override .npmrc so node-gyp runs
# for source-only packages (e.g. hnswlib-node@3.0.0). Keep this array in sync with Dockerfile:73.
```

CLAUDE.md
```diff
@@ -267,6 +267,7 @@ Project skills load from the `.claude/skills/` mount-point of the `skillsmith-st
 |---------|-----|
 | Container won't start | `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` |
 | Native module errors | `docker exec skillsmith-dev-1 npm rebuild better-sqlite3 onnxruntime-node` |
+| `hnswlib-node` fails validation after rebuild (`ignore-scripts=true` in `.npmrc` blocks node-gyp) | `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` — removes stale volume so Docker re-initialises from image layer (compiled without `.npmrc`). Permanent fix: SMI-5200 |
```
wrsmith108: **Apparent duplicate of the row above**

The prescribed fix — `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` — is character-for-character identical to the "Container won't start" row two lines up. A future editor may correctly identify them as duplicates and delete this row, losing the specific root-cause documentation.

Consider differentiating by explaining *why* the volume delete matters here vs. the generic "won't start" case, or by making the CLAUDE.md row point to the SMI-5200 permanent fix more prominently with the volume-delete command demoted to a parenthetical:

```suggestion
| `hnswlib-node` fails validation after rebuild (`ignore-scripts=true` in `.npmrc` blocks node-gyp) | Permanent fix ships in PR #1365 / SMI-5200. Until merged, delete the stale volume so Docker re-initialises from the image layer (compiled without `.npmrc`): `docker compose --profile dev down && docker volume rm skillsmith_node_modules && docker compose --profile dev up -d` |
```

docker-entrypoint.sh
wrsmith108: **Plausible: `--ignore-scripts=false` re-runs install hooks on `onnxruntime-node` and `esbuild`**

`npm rebuild <pkg> --ignore-scripts=false` overrides `.npmrc`'s `ignore-scripts=true` for **all** four modules — including `onnxruntime-node` and `esbuild`, which use network-fetched prebuilt binaries (not source compilation). Their `install` scripts attempt to download binaries from CDN.

In an air-gapped container or when the CDN is unavailable, `npm rebuild onnxruntime-node --ignore-scripts=false` will fail trying to re-fetch a binary that is already present and working. `|| echo` prevents an abort, but the rebuild loop then marks the module as "exited non-zero" and the re-validation re-check at line 143 may find the module actually still loads — which is correct — but the noise and delay are unnecessary.

This flag was added specifically for `hnswlib-node` (the only source-only module). Scoping it to only source-compiled modules avoids triggering network operations on the others:

```suggestion
        if [ "${module}" = "hnswlib-node" ]; then
          npm rebuild "${module}" --ignore-scripts=false || echo -e "${YELLOW}  ↳ npm rebuild exited non-zero for ${module} (see output above)${NC}"
        else
          npm rebuild "${module}" || echo -e "${YELLOW}  ↳ npm rebuild exited non-zero for ${module} (see output above)${NC}"
        fi
```

Or, more cleanly, a lookup table of source-compiled modules if the list grows.

PR #1542, opened 2026-06-22
fix(indexer): remove created:> date qualifier from Deno code search (SMI-5176)

hytonylee:
## Problem

The daily skill indexer's Phase 3a and 3b legs — which use GitHub's code search API to discover \`SKILL.md\` files across all public repositories — were silently returning ~0 results on every run.

The consequence: the Skillsmith registry stopped picking up newly published skills. Any skill created or updated after the freshness window would never appear in search results, recommendations, or the MCP tools, even though it existed publicly on GitHub.

Root cause: the indexer was passing \`created:>2026-06-15\` as a qualifier to \`GET /search/code\`. GitHub's code search API does **not** support \`created:\` as a date filter — it tokenizes it as a literal content term, matching only files whose text contains that date string. A direct API comparison confirmed the scope of the breakage:

\`\`\`
# Without qualifier (correct) → 9,096 results
GET /search/code?q=filename:SKILL.md+path:/

# With qualifier (broken) → 0 results
GET /search/code?q=filename:SKILL.md+path:/+created:>2026-06-15
\`\`\`

The Node/scripts indexer was already fixed under SMI-5176. This PR aligns the Deno edge function to match.

## Solution

- Removes \`createdAfter\` from \`searchCodeForSkillMd\` and \`searchCodeForSkillMdInSubdirectory\` in the Deno edge function and stops passing \`freshnessDate\` from the Phase 3a/3b callers
- \`freshnessDate\` continues to flow into Phase 2 topic/repo search (\`searchRepositories\`) where \`created:>\` **is** a valid filter and works correctly

## Test plan

- [x] TypeScript typecheck passes (\`npm run typecheck\`)
- [x] Indexer unit/contract tests pass (58 tests across \`indexer-runners\`, \`index.subdirectory\`, \`discovery-orchestrator\`)
- [ ] Post-deploy: verify \`phase3a.repos_found\` in \`audit_logs\` jumps from ~0 to hundreds on next indexer cron run

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wrsmith108:
Closing this without merging — cross-checked against Linear and this fix targets code with no live production path.

**SMI-5176** (the issue this PR's title/branch reference) is already **Done**, completed 2026-06-18 via PR #1482. That fix deliberately scoped to the Node `scripts/indexer/` tree only — the authoritative indexer the production cron (`indexer.yml`) actually runs. The Deno copy under `supabase/functions/indexer/` was intentionally left untouched (see the SMI-5176 comment thread: "Deno copy untouched (SMI-5182; not parity-guarded — the issue's 'Deno copy only' note was wrong)").

**SMI-5182** ("Tech-debt — decide fate of legacy `supabase/functions/indexer/` Deno copy") is the tracker for whether this copy gets fixed-and-parity-guarded or deleted outright. It's still open, and the latest update (2026-07-04) found the deployed Deno indexer function is now unreachable via any known invocation path in prod (its pg_cron trigger was dropped in migration `060_drop_invoke_indexer.sql`, and `indexer-dispatch` routes to the GHA/Node workflow instead) — so the lean is toward deletion, not toward maintaining parity fixes like this one.

Merging this would resurrect exactly the dual-edit maintenance hazard SMI-5182 flags, on code with no prod runtime path. If/when SMI-5182 resolves in favor of keeping the Deno copy alive, this diff is still valid and can be revived from this branch — but landing it now would work against the pending decision. Thanks for the fix regardless — the root-cause diagnosis (GitHub's code search tokenizing `created:>` as content) was correct and is exactly what the Node-side fix addressed.

PR #1710, opened 2026-07-04
fix(security): guard get_user_subscription IDOR + revoke anon (SMI-5526)

wrsmith108:
## SMI-5526 (expedited): get_user_subscription IDOR fix

`get_user_subscription(uuid)` was `SECURITY DEFINER` with no `auth.uid()` guard and an `anon` EXECUTE grant — any caller could read any user's subscription (tier/status/seats/period) via `/rest/v1/rpc/get_user_subscription`, bypassing RLS. Same class as SMI-5525, lower severity (billing metadata).

**Fix**: add an `auth.uid() <> user_uuid` guard (transparent to the website account pages, which pass their own `session.user.id`) + revoke `anon`/`PUBLIC`, keep `authenticated`. Body reproduced verbatim from prod `pg_get_functiondef`.

**Already applied to prod** (`vrcnzpmndtroqxxoqkzy`, recorded `20260703235000`) and staging. Anon `/rpc/get_user_subscription` now returns `401 permission denied`. This PR lands the migration file for repo parity — do not re-apply.

Plan: `docs/internal/implementation/smi-5526-definer-grant-audit.md`.

`[skip-impl-check]` encrypted migration, no counted source change. `[skip-smoke]` prod already applied + verified.


wrsmith108:
Consolidated into #1711 after a sibling migration-number collision forced a renumber (20260704000000/000001). Prod already reflects the expedite order (IDOR applied first).

PR #1897, opened 2026-07-14
docs(internal): Bump docs/internal pointer for SMI-3140 confirmed decisions

wrsmith108:
## Summary

Bumps `docs/internal` to include the SMI-3140 plan's three confirmed decisions (sparse-data default, extraction-hardening now in-scope for Wave 1 as Step 3b, no-redaction default) — see skillsmith-docs#423.

## Test plan

- [x] Docs-only change (submodule pointer + no other files)
- [x] Preflight/security/format/test checks passed via this worktree's pre-push hooks

Refs SMI-3140, SMI-5676

[skip-impl-check]


wrsmith108:
Redundant: outer main's docs/internal gitlink already points at 4e8783a, which is a descendant of this PR's target commit 662dfd5 (verified via git merge-base --is-ancestor). The three-confirmed-decisions plan update is already live on main. No content lost — closing without merging to avoid regressing the pointer backward.

Issue #1912, opened 2026-07-15
Cross-harness inventory hides symlinked harness membership


caioribeiroclw-pixel:
## Problem

The cross-machine inventory is positioned as a per-harness audit, but a skill directory symlinked into more than one supported harness is reported under only the first harness in `CLIENT_IDS`.

`collectDeviceSkills()` shares one `seenRealpaths` set across all harness scans. `collectHarness()` skips any realpath already seen, so with the current order a skill physically stored under `~/.claude/skills/foo` and symlinked at `~/.cursor/skills/foo` produces only a `claude-code` row.

The behavior is explicit in:

- `packages/core/src/sync/inventory-collector.ts` ("first harness in CLIENT_IDS order wins")
- `packages/core/src/sync/inventory-collector.test.ts`, IC-2

## Why this matters

For a cross-harness/compliance view, physical deduplication and harness membership are different facts. The same bytes may intentionally be reused by multiple harnesses. Collapsing the alias makes the dashboard imply that the skill is absent from Cursor even though Cursor can discover it.

This is especially easy to hit when users keep one canonical skill directory and symlink it into Claude Code, Cursor, Copilot, or the shared `~/.agents/skills` path to avoid divergent copies.

## Minimal reproducer

```bash
mkdir -p ~/.claude/skills/foo ~/.cursor/skills
printf '%s\n' '---' 'name: foo' 'version: 1.0.0' '---' > ~/.claude/skills/foo/SKILL.md
ln -s ~/.claude/skills/foo ~/.cursor/skills/foo
skillsmith inventory status --verbose
skillsmith inventory push
```

The collector emits one `foo` entry for `claude-code`; no `cursor` membership is uploaded.

## Expected product semantics

Keep one content identity/hash if desired, but preserve every observed harness membership. Two possible representations:

1. emit one row per `(harness, skill_id)` while deduplicating hash computation by realpath; or
2. represent one physical skill with a `harnesses[]` membership list and render it as shared.

At minimum, the dashboard should not label the later harness as missing when its skill path resolves to the same directory.

This surfaced while reviewing the fresh cross-machine inventory UAT request in the Cursor community. I have not enrolled as a design partner or tested the hosted dashboard; this report is based on the current scanner and its tests.

wrsmith108:
Thanks for the clear, well-diagnosed report, @caioribeiroclw-pixel — the exact repro (symlinking a skill directory into two harnesses) made this easy to track down and verify.

[...]

PR #1913, opened 2026-07-15
fix: preserve cross-harness skill membership

caioribeiroclw-pixel:
## Summary

- keep realpath deduplication within each harness
- preserve a separate inventory row when the same physical skill is discoverable by another harness
- update IC-2 to assert both harness memberships share one content hash

This addresses #1912. The inventory primary key and product claim are per `(harness, skill_id)`, so physical deduplication should not erase a destination edge.

## Verification

- `npx vitest run packages/core/src/sync/inventory-collector.test.ts` — 7/7
- `npx eslint packages/core/src/sync/inventory-collector.ts packages/core/src/sync/inventory-collector.test.ts`
- `npx tsc --noEmit --pretty false -p packages/core/tsconfig.json`
- `git diff --check`

### DIFF
--- a/packages/core/src/sync/inventory-collector.test.ts
+++ b/packages/core/src/sync/inventory-collector.test.ts
@@ -7,7 +7,7 @@
  * without re-importing the module.
  *
  * IC-1: same skill under two harnesses (distinct realpaths) -> two entries.
- * IC-2: a symlinked alias across harnesses -> one entry (first ClientId wins).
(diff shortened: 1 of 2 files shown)


caioribeiroclw-pixel:
Thanks — confirmed #1923 covers the original collector semantics and the additional CLI path the literal symlink repro exposed. It preserves one row per harness, keeps within-harness realpath dedup, and adds the individually-symlinked-directory discovery case. Closing this narrower patch as superseded by #1923.

PR #2068, opened 2026-07-26
chore(eval): cron-detected baseline drift (2026-07-26)

wrsmith108:
Auto-generated by `scripts/eval-baseline-cron.sh` in the isolated eval checkout. Review the Retrieval Eval Gate per-category breakdown before merging.

baseline.json sha256: `1669077816de` -> `25ee4136ef3c`

[skip-impl-check]


wrsmith108:
`Retrieval Eval Gate` is failing on a pre-existing CI infra bug unrelated to this PR's diff (only touches `baseline.json`/`.signatures.log`/`.cron-heartbeat`) — the job hardcodes `npx -p vitest@4.1.2`, which no longer matches root's pinned vitest (`4.1.10`), and running a mismatched version via `npx -p` inside a workspace member directory triggers a known npm Arborist null-pointer crash (`edgesOut`). Reproduced on 2 consecutive CI runs. Filed as SMI-5931 with root cause + fix options.

PR #2192, opened 2026-08-05
feat(indexer): SMI-5879 Wave 2 — RC-1 fix + edge-twin scanner parity port

wrsmith108:
## Business Summary

**What shipped:**
- Fixed a bug where the security scanner only checked the *first* occurrence of certain jailbreak/prompt-injection phrases across an entire skill document instead of every occurrence — a real attack repeated later in a long document could undercount and slip past the quarantine threshold.
- Ported the same detection patterns and scoring model from the main scanner to the production edge scanner, so both engines now agree on what counts as a jailbreak attempt or prompt injection, and at what severity.
- Added a safety check ("Gate F") so the live production indexer refuses to run during a declared safety freeze window — closing the one path an outside caller (or a public API key holder) could still trigger a run during a freeze.
- Tightened the scanner's fast pre-check so a page merely *mentioning* "jailbreak" or "developer mode" no longer fails it on its own — only genuine directive/attack language does.
- Added the design's required proof fixture: the same realistic document that used to score clean now correctly quarantines under the fix, with the exact category-by-category math asserted, not just a pass/fail.

**Quality bar:** Full repo-wide test suite verified after every change (15,702 tests, 0 failures), including a dedicated 98-case regex-timing safety battery (all patterns complete in well under 1% of their budget) and a new automatic guard that fails loudly — instead of silently skipping — if the two copies of the scanner code (used by two different runtimes) ever drift out of sync.

**Found along the way:** three lower-priority pattern-coverage gaps identified during the port, filed as separate follow-ups rather than bundled into this fix: [SMI-5923](https://linear.app/smith-horn-group/issue/SMI-5923) (5 additional jailbreak phrasings, medium priority), [SMI-5924](https://linear.app/smith-horn-group/issue/SMI-5924) (2 chat-transcript-style injection formats, medium priority), [SMI-5925](https://linear.app/smith-horn-group/issue/SMI-5925) (a code-comment accuracy cleanup, low priority).

**Net result:** Implementation, tests, and CI are ready — but **this PR must not be merged yet**. Per the design's own safety window, merge requires the freeze-detection gate to first run inertly through one full day-and-night operational cycle (no earlier than **2026-08-06 01:18 UTC**), plus a live confirmation that the freeze gate actually blocks a real run when engaged. Both are tracked in SMI-5879; merging early would risk deploying unverified code into the production quarantine gate that runs against 344,716+ skill rows.

---

### RC-1 — multiline-scan first-match-only bug (core)
`scanPatternsWithMultilineSupport` (`packages/core/src/security/scanner/SecurityScanner.helpers.ts`) tested each 'content'/'both'-scope pattern via a non-global regex `.match()`, returning only the first match in the whole document. Fixed with a per-pattern global-exec loop bounded by two independent constants: `MAX_MULTILINE_LINES_PER_PATTERN = 64` (score-neutral — proven in-code, Lemma 3.3-B) and `MAX_MULTILINE_ITERATIONS_PER_PATTERN = 10_000` (wall-clock liveness only, sets `ScanReport.multilineTruncated`). Also fixed a genuine off-by-one found via testing: a match capturing a leading `(?:^|\n)` anchor reports `.index` at the *previous* line's newline, not the matched line's own start.

Also raised `MAX_CONTENT_LENGTH_FOR_REGEX` (pass-1 full-content cap) from 10,000 to 1,000,000 chars, closing a truncation blind spot on real-sized SKILL.md files — independently re-benchmarked (not trusting an earlier design-doc estimate): worst case ~56ms at 1MB, confirmed linear (not exponential) to 4MB.

### Edge scanner architectural port
- `JAILBREAK_PATTERNS` grows 10→18 entries, `PROMPT_INJECTION_PATTERNS` 6→9, in both `scripts/indexer/_shared/security-scanner-edge.patterns.ts` and its Deno twin `supabase/functions/_shared/`.
- New evidence-tier classification + pattern-scope + corroboration module (`security-scanner-edge.evidence.ts`) and two-pass multiline-scan engine (`security-scanner-edge.multiline.ts`), ported from core's already-reviewed `SecurityScanner.evidence.ts`/`patterns.jailbreak.evidence.ts`/`patterns.scope.ts`/fixed `scanPatternsWithMultilineSupport`.
- `scanJailbreakPatterns`/`scanPromptInjection` rewritten onto the new engine; `escalateCorroboratedMentions` wired in after `escalateCodeExecution`.
- `quickCheck`/`quickSecurityCheck` (both core and edge) now test only the directive-tier derived subset of `JAILBREAK_PATTERNS` (design §5) — a bare mention no longer fails the fast pre-filter.
- `security-scanner-edge.ts` split into `.regex-utils.ts` and `.chmod-compound.ts` siblings to stay under the file-size convention.

### Gate F — pre-writer freeze gate (`supabase/functions/indexer/`)
New `freeze-gate.ts`: a Deno-side reimplementation (not a byte-identical twin — the Node-side `run-gate.ts` must never be mirrored under `supabase/functions/`, since one file there converts a gate-infra PR into a 31-function deploy fanout) of the same freeze-marker vocabulary. Called in `index.ts` right after the method check, before the row-level lock RPC and before any dispatch — refuses (HTTP 503) and records a distinctly-event-typed audit row when the marker's allowlist doesn't permit the request's run type.

### SW-1 witness fixture (design §3.5)
`packages/core/tests/security/scanner-witness-sw1.test.ts`: a single realistic document with a fully specified category budget proving the verdict swing concretely — baseline (frozen, verbatim pre-fix logic) `riskScore=37` (clean); ported (current) `riskScore=42` (quarantine). Per-category breakdown is asserted, not just the total.

### Twin parity + regex safety (design §6/§7)
- `scripts/tests/indexer/parity.test.ts` extended with whole-body + per-function parity checks for the 4 new sibling modules, plus a new hard-fail guard (`SKILLSMITH_PARITY_ALLOW_ENCRYPTED_DENO=1` opt-out, registered in `docs/internal/process/guards-and-opt-outs.md`) so an encrypted Deno twin fails loudly instead of silently skipping its parity assertions.
- `security-scanner-edge.regex-safety.test.ts`: table-driven adversarial timing battery (4 shapes × 4 sizes × 6 patterns), 3-warmup + median-of-5 protocol, budgets 50ms/250ms/2000ms. Measured worst case 0.29ms per-pattern, 49ms whole-scan on a 1MB bundle.

### CI status
Full repo-wide `npx vitest run`: 15,702 tests passed, 0 failed (verified after every commit, and again after a rebase onto current `main`). `npm run audit:standards`: 0 failures. Typecheck/lint/format: clean.

### Do not merge yet
Merge is gated on the design's own P-0 pre-conditions (§8.3.2.0): the freeze gate must be observed inert across a full cron calendar with `INDEXER_RUN_ALLOWLIST` unset, positive controls must fire red under an engaged freeze, and ≥24h must elapse from the retro-fix merge that made the gate's env-plumbing effective (`4e831bb2a`, 2026-08-05T01:18:28Z) — so not before **2026-08-06T01:18 UTC**. The actual merge/deploy is an attended ~13-hour operational runbook (design §8.3.2.5.6), not a normal squash-merge, and needs explicit scheduling.


wrsmith108:
## Merge proceeding with an explicit runbook-gate override

This PR's own body states it must not merge before the attended production runbook (design §8.3.2.5.6) completes, gated on `smi5879-gate-check.ts`'s G-1/G-2. As of this merge:

- **The real 3-shard `--purpose=decision` production dispatch has not run.** What ran instead was a single-process scanner-parity simulation (`smi5879-simulate-full.ts`) over the already-sealed decision-purpose census (SMI-6015 Wave 3 Step 1, sealed 2026-08-31, 473,135 rows).
- That simulation reached a terminal state on 2026-09-07: `hard_stopped=max_passes` (8-pass ceiling), not full convergence. Residual: 33,040 → 28,449 across the run, never reaching zero. All 4 cohorts report `partial` coverage (not `full`) — C1–C4 unevaluable counts: 90 / 10,661 / 8,044 / 9,654; unfetchable: 37 / 7,044 / 6,355 / 9,641.
- Per `smi5879-gate-check.gates.ts`, **Gate G-2 (coverage) fails on this result by design** — `hard_stopped != null` and no cohort is `full`. Gate G-1 (final sign-off) cannot even be evaluated until G-2 passes.

Ryan reviewed this exact finding and explicitly decided to merge anyway, overriding G-2, rather than holding for investigation or a re-run. This is a deliberate, informed decision — not a default or an inferred approval — made after seeing the specific failure mode (non-convergence hard-stop, ~28K stuck rows, largely `unfetchable`/`bundle_absent`-adjacent) and skillsmith-dd's assessment that this may reflect a real data-quality issue rather than transient flakiness.

Filed for follow-up, not silently dropped: the ~28K rows that never resolved should still be investigated (SMI-6015), independent of this merge decision.

PR #2206, opened 2026-08-06
feat(mcp): Detect and notify on live MCP disconnect (SMI-5941)

wrsmith108:
## Business Summary

**What shipped:** the `skillsmith`/`skillsmith-doc-retrieval` MCP servers can now be trusted to tell you when they've disconnected mid-session — previously this could happen silently, with tools quietly vanishing while the Docker container stayed perfectly healthy the whole time, and no way to know except stumbling onto it by accident (this is exactly what happened during the investigation that led to this PR). Two independent detection paths now cover the two ways this can happen: an immediate, harness-guaranteed warning when a tool call is actually attempted against a broken connection, and a best-effort assistant nudge for the case where the tool disappears before any call is even tried. Either way, you get told which server dropped and the exact steps to reconnect, both in the moment and again at your next session if you missed it.

**Quality bar:** implementation went through three rounds of independent cross-model review (a different AI model family, specifically to catch blind spots a same-family review might share) — each round caught and fixed a real defect before it could ship: the first design targeted the wrong underlying event entirely and would never have fired; the second design's protection against data races wasn't actually race-proof; the third round caught a subtle bug in that same protection where a slow-but-still-working process could have its safety lock incorrectly stolen out from under it. All three are fixed and independently re-verified. Full test suite (602 tests) passes, including new tests that exercise the fix by spawning real, separate processes to force the exact race conditions being guarded against.

**Found along the way:** a real bug in this repo's own internal plan-review tooling (`scan-plan.sh`) was silently skipping most of a plan document's content during review — filed and fixed separately (SMI-5942, [skillsmith-strategy#13](https://github.com/smith-horn/skillsmith-strategy/pull/13)) rather than bundled here.

**Net result:** fully shipped for both detection paths. The paired [docs/internal PR](https://github.com/smith-horn/skillsmith-docs/pull/643) carries the full design write-up and review history. Not included: automatically fixing a disconnected server on your behalf — reconnecting is still a manual `/mcp` step, since there's no way to trigger that programmatically today.

---

## Technical detail

**Track A** (ships independently, no infra-change gate): `CLAUDE.md` — new instruction telling any session to treat an MCP-disconnect system-reminder as always worth surfacing immediately, rather than as skippable background context.

**Track B** (the gated infra change):
- `packages/doc-retrieval-mcp/src/retrieval-log/mcp-disconnect-state.ts` (+ `.test.ts`) — new shared state module: records/acknowledges disconnects, guarded by a portable mkdir-based mutual-exclusion lock with PID-liveness-verified stale-lock reclaim (a first for this codebase's state-file family — none of the existing siblings needed real concurrent-writer protection).
- `packages/doc-retrieval-mcp/src/_lib/mcp-disconnect-worker.ts` — test-only helper process for the real-concurrency test suite.
- `scripts/mcp-disconnect-state.ts` — thin CLI wrapper over the shared module.
- `scripts/session-mcp-disconnect-guard.sh` (+ `scripts/tests/...test.sh`) — the new `PostToolUseFailure` hook, wired into `.claude/settings.json`.
- `scripts/session-priming-query.ts` — new fourth `SessionStart` banner, alongside the existing auto-heal/liveness/reindex banners.
- `.github/workflows/validate-hooks.yml` — registers the new hook + test in the existing host-only hooks CI job (Claude Code hooks run on the host, never inside the Docker dev container).
- `packages/doc-retrieval-mcp/CHANGELOG.md` — `[Unreleased]` entry.

Verification: `npm run typecheck` / `npm run lint` / `npm run audit:standards` (0 failures) all clean; `npx vitest run packages/doc-retrieval-mcp` — 602/602 passing; `bash scripts/tests/session-mcp-disconnect-guard.test.sh` (run on host, not in Docker — see the test file's header for why) — 28/28 passing.

Follow-up (tracked, not blocking): once the paired `docs/internal` and `skillsmith-strategy` PRs merge, a separate `chore(docs): bump pointer` commit will update this repo's submodule references.

Refs SMI-5941.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -159,6 +159,18 @@
         ]
       }
     ],
+    "PostToolUseFailure": [
+      {
+        "matcher": "mcp__skillsmith",
+        "hooks": [
+          {
+            "type": "command",
+            "timeout": 5,
+            "command": "$CLAUDE_PROJECT_DIR/scripts/session-mcp-disconnect-guard.sh"
+          }
+        ]
+      }
+    ],
     "PreCompact": [
       {
         "matcher": "manual",
--- a/.github/workflows/validate-hooks.yml
+++ b/.github/workflows/validate-hooks.yml
@@ -30,6 +30,8 @@ on:
       - 'scripts/worktree-docker.sh'
       - 'scripts/docker-health.sh'
       - 'scripts/session-start-mcp-command-guard.sh'
+      - 'scripts/session-mcp-disconnect-guard.sh'
+      - 'scripts/tests/session-mcp-disconnect-guard.test.sh'
       - 'scripts/tests/create-worktree-hooks.test.sh'
       - 'scripts/tests/session-start-audit.test.sh'
       - 'scripts/tests/reclaim-node-modules.test.sh'
@@ -80,6 +82,8 @@ jobs:
             scripts/worktree-docker.sh \
             scripts/docker-health.sh \
             scripts/session-start-mcp-command-guard.sh \
+            scripts/session-mcp-disconnect-guard.sh \
+            scripts/tests/session-mcp-disconnect-guard.test.sh \
             scripts/tests/create-worktree-hooks.test.sh \
             scripts/tests/session-start-audit.test.sh \
             scripts/tests/reclaim-node-modules.test.sh \
@@ -112,6 +116,8 @@ jobs:
           bash -n scripts/worktree-docker.sh
           bash -n scripts/docker-health.sh
           bash -n scripts/session-start-mcp-command-guard.sh
+          bash -n scripts/session-mcp-disconnect-guard.sh
+          bash -n scripts/tests/session-mcp-disconnect-guard.test.sh
           bash -n scripts/tests/create-worktree-hooks.test.sh
           bash -n scripts/tests/session-start-audit.test.sh
           bash -n scripts/tests/reclaim-node-modules.test.sh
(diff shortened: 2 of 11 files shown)


wrsmith108:
## pr-reviewer results (gpt-5.6-sol via NEEDLE, ADR-128)

**Verdict: PASS-WITH-FIXES** — ready to merge once CI is green.

| Check | Result |
|---|---|
| PR-01–06, 08, 09, 11, 13 | SKIP (no SQL/edge-function/enum/column triggers in this diff) |
| PR-07 (silent catch) | PASS |
| PR-10 (export continuity) | SKIP (no renames/moves) |
| PR-12 (plan vs diff) | FAIL→FIXED — plan's Step 6 test coverage was incomplete at the bash guard-script layer (no test proved `systemMessage` still fires under a real lock timeout, and a reciprocal namespace-isolation assertion was missing); fixed in 479587ea9, bash suite now 32/32 |
| PR-14 (audit:standards) | PASS (couldn't run in the Codex sandbox — Docker socket denied — ran manually: 0 failures) |

Full report: `docs/internal/pr-reviews/2026-08-06-smi-5941-mcp-disconnect-detection.md` (skillsmith-docs#643).

Unrelated: a brand-new high-severity `js-yaml` CVE started failing the repo-wide pre-push audit gate mid-review — filed as SMI-5943, one `--no-verify` push used with explicit approval to avoid blocking this PR on it.

PR #2308, opened 2026-08-11
fix(core): Make sql.js persist() atomic and detect empty DBs

wrsmith108:
## Business Summary

**What shipped:** the local Skillsmith database (`~/.skillsmith/skills.db`) can no longer be silently corrupted into a 0-byte file by an interrupted write, and if it ever does end up empty for another reason, the server now fails with a clear, actionable error instead of crashing opaquely.

**Quality bar:** full `packages/core` suite (5,388 tests) green, plus two new regression tests reproducing the exact incident (an interrupted write leaving the file untouched, and a 0-byte file producing a clear error instead of a migration crash). Lint, typecheck, and `audit:standards` all clean.

**Found along the way:** a pre-existing flaky test (`crash-handler-integration.test.ts`, SMI-5787) — a spawn-based integration test with a hardcoded 15s timeout that races under concurrent host load — filed separately as SMI-5999 rather than touched here, since it's unrelated to this change.

**Net result:** fully shipped, no follow-up on this PR itself. SMI-5999 tracks the unrelated flaky test found while pushing.

---

## Technical detail

**Root cause of the incident:** `~/.skillsmith/skills.db` was found truncated to 0 bytes, causing the `skillsmith` MCP server to fail every reconnect with `CONNECTION_CLOSED` on this machine (all sessions). Two compounding bugs in `@skillsmith/core`'s WASM (sql.js) database layer:

1. `SqlJsDatabaseAdapter.persist()` (`packages/core/src/db/drivers/sqljsDriver.ts`) wrote the exported DB buffer via a single `writeFileSync()` call to the live file path — not atomic. A process kill mid-write (OOM, SIGKILL, machine sleep) truncates the file before the new bytes land.
2. `openDatabaseAsync()` (`packages/core/src/db/schema.ts`) treated any file missing a `schema_version` table as a legacy import, silently stamping `schema_version=1` and running every migration without ever creating the base schema. Against an empty file this failed opaquely (`no such table: skills`, ..., fatally `no such table: cache`), crashing server startup with no diagnostic surfaced to the MCP host.

**Fix:**
- `persist()` now writes to a temp file (`<path>.tmp`) and `renameSync`s it over the target — rename is atomic on the same filesystem, so the target file is always either fully-old or fully-new, never truncated.
- `openDatabaseAsync()` now distinguishes a genuinely empty/corrupt database (zero tables at all) from a real legacy import (has tables, just missing `schema_version`) — the former throws an actionable error with remediation instead of silently misclassifying it and crashing later.

**Tests added:** `packages/core/tests/db/sqljsDriver.test.ts` (atomic-write + interrupted-write-leaves-original-untouched), `packages/core/tests/db/schema-async.test.ts` (0-byte file throws actionable error; legacy-import path still works).

**CI caveat:** pushed with `--no-verify` after explicit user sign-off — the pre-push hook's only failure was the unrelated flaky SMI-5787 test (confirmed via isolated run: passes reliably in 6-7s; only times out in the full suite while two other tests in the same run consume 85s/117s under heavy concurrent load from 16 sibling Docker containers on this machine). CI on this PR will run cleanly regardless of host contention.

Fixes SMI-5997

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to `@skillsmith/core` are documented here.
 ## [Unreleased]
 
 - **Fix**: new shared `extractContextWords()` (`services/context-words.ts`, exported from the package root) replaces a `.filter((w) => w.length > 3)` threshold both `@skillsmith/mcp-server`'s `skill_recommend` and `@skillsmith/cli`'s `recommend --context` used independently — it was silently dropping real short technical terms ("git", "ci", "aws", "sql", "k8s") from the recommendation stack, tripping the empty-stack guard even when usable context was supplied (SMI-5986)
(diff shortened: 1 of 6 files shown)


wrsmith108:
Closing and reopening to force GitHub to resync check-suites — the GitHub Actions check-suite never registered for the last two commits (only Claude/Vercel app check-suites appeared) despite CI running normally elsewhere in the repo at the same time.

PR #2358, opened 2026-08-14
feat: add code-health-auditor skill (dead-code/consolidation scanner)

wrsmith108:
## Business Summary

**What shipped:** A new `code-health-auditor` skill that scans the monorepo for dead code and consolidation candidates during code review — it cross-references every candidate against real test coverage before suggesting anything, and never flags code a live test still needs. It never auto-deletes anything; it produces a report for a human to act on.

**Quality bar:** The design was reviewed by an independent model (GPT-5.6-Sol) before any code was written, which found 13 real problems with the first draft and required all of them fixed before implementation started — including a coverage check that was too coarse (whole-file instead of the exact flagged code) and a heuristic for Supabase function liveness that turned out to be wrong 18% of the time when checked against this actual repo. The finished tool was verified end-to-end with a real run against one package, producing a correct four-category report.

**Found along the way:** a pre-existing high-severity dependency vulnerability (`nanoid`, unrelated to this change) that this PR's own security check caught — fixed as a small drive-by patch rather than filed separately, since the fix was a one-line override.

**Net result:** Fully shipped for one package (`packages/cli`), which is the only package this tool currently trusts for its strongest "safe to delete" verdict. Every other package is explicitly marked experimental and downgraded to "needs more evidence" until it gets the same calibration — a deliberate scope boundary, not a gap that was missed. Applying the tool against the rest of the repo and writing up the results (Wave 2), plus a blog case study (Wave 3), are separate follow-up work.

---

## Technical detail

**Companion PRs** (this repo uses submodules for the skill itself and internal docs):
- Skill files: [smith-horn/skillsmith-strategy#15](https://github.com/smith-horn/skillsmith-strategy/pull/15) — `.claude/skills/code-health-auditor/{SKILL.md, patterns/README.md, patterns/calibration/*, scripts/scan-repo.sh, scripts/verify-candidates.sh}`
- Plan doc + guard registration: [smith-horn/skillsmith-docs#755](https://github.com/smith-horn/skillsmith-docs/pull/755) — `docs/internal/implementation/code-health-auditor.md` (includes the full GPT-5.6-Sol review record), `docs/internal/process/guards-and-opt-outs.md`

**This PR** (outer repo):
- `package.json` / `package-lock.json` — add `knip` `6.32.2` (exact-pinned) as a root devDependency.
- `knip.json` (new) — per-workspace config; `packages/cli` is calibrated (real Knip output checked into the companion skill PR as fixtures), every other `packages/*` entry is commented `EXPERIMENTAL` and the skill's own `verify-candidates.sh` hard-downgrades its findings regardless of coverage result.
- `package.json` `overrides.nanoid` → `^3.3.18` + regenerated lockfile — fixes a pre-existing high-severity CVE (`GHSA-2v37-7h3g-55p8`, nanoid's zero-size generator infinite loop), a transitive dep of `postcss` already on `main`, unrelated to the skill itself. Regenerated via `npm install --package-lock-only --ignore-scripts` (this repo's established pattern for adding a lockfile entry without touching native modules).
- `.claude/skills`, `docs/internal` — submodule pointer bumps to the companion PRs' branch tips.

**Design** (see the plan doc in the docs companion PR for full detail): the skill runs Knip in two modes per workspace (with tests as entry points, and Knip's own `--production` mode) plus a small grep pass for string-dispatched code Knip's import graph can't see. Its safety gate (`verify-candidates.sh`) maps every Knip finding to the exact enclosing function/statement range in a fresh Vitest coverage report — not a whole-file aggregate — and treats "no coverage data" as strictly weaker evidence than "confirmed zero coverage." Output is four buckets: Safe to delete, Consolidation candidate, Looks bad but is fine (split into real false positives vs. known blind spots like Supabase edge functions), and Needs runtime verification for anything the gate can't confirm either way.

**Verification**: `scan-repo.sh --package cli` → `verify-candidates.sh` run live inside the worktree container against real `packages/cli` source and a fresh coverage run, producing a correct report (7 Safe-to-delete, several Consolidation candidates, both Looks-bad-but-fine subsections populated). `SKILLSMITH_CODE_HEALTH_AUDIT_DISABLE=1` verified to short-circuit both scripts. `shellcheck -S warning` clean. Local `lint` / `typecheck` / `audit:standards` / `format:check` all pass.

**Known pre-existing gap, not from this PR**: `npm run preflight`'s dependency-import scanner reports 3 missing packages (`@isaacs/keytar`, `async_hooks`, `@wdio/globals`) in files this PR never touches (`packages/core/src/config/*`, `packages/vscode-extension/e2e/*`) — confirmed via `git diff main` that this PR has zero overlap with those files.

Refs SMI-6023

`[skip-impl-check]` — this repo's `verify-implementation` CI check only sees this repo's own diff (`package.json`/`package-lock.json`/`knip.json` + submodule gitlink pointer bumps), not the actual skill implementation (bash scripts, docs), which lives in the two companion submodule PRs (smith-horn/skillsmith-strategy#15, smith-horn/skillsmith-docs#755) that this PR's gitlink bumps point at. Real implementation exists — it's just not visible to a diff scoped to this repo alone.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -1,0 +1,55 @@
+{
+  "$schema": "https://unpkg.com/knip@6/schema.json",
+  "workspaces": {
+    "packages/cli": {
+      "entry": ["src/index.ts", "src/**/*.test.ts", "tests/**/*.ts"],
+      "project": ["src/**/*.ts", "tests/**/*.ts"],
+      "ignoreDependencies": [
+        "@opentelemetry/instrumentation-http",
+        "@opentelemetry/instrumentation-runtime-node",
+        "@opentelemetry/instrumentation-undici",
+        "@opentelemetry/resources",
+        "@opentelemetry/sdk-node",
+        "@opentelemetry/sdk-trace-base",
+        "@opentelemetry/semantic-conventions",
+        "fts5-sql-bundle",
+        "posthog-node",
+        "tree-sitter-wasms",
+        "web-tree-sitter",
+        "yaml"
+      ]
+    },
+    "packages/core": {
+      "entry": ["src/index.ts", "src/**/*.test.ts", "tests/**/*.ts"],
+      "project": ["src/**/*.ts", "tests/**/*.ts"]
+    },
+    "packages/mcp-server": {
+      "entry": ["src/index.ts", "src/**/*.test.ts", "tests/**/*.ts"],
+      "project": ["src/**/*.ts", "tests/**/*.ts"]
+    },
+    "packages/enterprise": {
+      "entry": ["src/index.ts", "src/**/*.test.ts", "tests/**/*.ts"],
+      "project": ["src/**/*.ts", "tests/**/*.ts"]
+    },
+    "packages/doc-retrieval-mcp": {
+      "entry": ["src/**/*.ts", "src/**/*.test.ts"],
+      "project": ["src/**/*.ts"]
+    },
+    "packages/vscode-extension": {
+      "entry": ["src/extension.ts", "src/**/*.test.ts"],
+      "project": ["src/**/*.ts"]
+    },
+    "packages/skillsmith-cli": {
+      "entry": ["src/**/*.ts"],
+      "project": ["src/**/*.ts"]
+    },
+    "packages/billing-types": {
+      "entry": ["src/index.ts"],
+      "project": ["src/**/*.ts"]
+    },
+    "packages/website": {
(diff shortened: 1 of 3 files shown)


wrsmith108:
Closing and reopening to force a genuinely fresh CI run (diagnostic — investigating non-deterministic npm ci failures, not a content issue)

PR #2396, opened 2026-08-17
fix(SMI-6050): worktree containers now get real Linux build-tool binaries

wrsmith108:
## Business Summary

**What shipped:** Worktree Docker containers now correctly get every build tool's Linux-platform binary (turbo, Rollup/Rolldown, Astro's compiler, Lightning CSS, Tailwind's Oxide engine, ruvector, Wrangler/workerd) — previously these were silently missing because the shared host `node_modules` tree is built on macOS, which skips Linux-only optional dependencies, and worktree containers only ever saw that host tree read-only. This caused real build failures (`astro check` / "Cannot find native binding") three times in one prior session, each worked around by hand.

**Quality bar:** SPARC research + an independent GPT-5.6-Sol plan-review before any code was written (5 blocking issues found and fixed in the plan itself — an undefined seed-to-target mapping, missing version-staleness detection, a factually wrong exclusion of Wrangler/workerd, a Dockerfile layer-ordering bug, and an unsafe rollout order). Implemented across 4 sequential waves, each independently verified against a real running container — not just unit tests. The actual production bug was reproduced before the fix and confirmed fixed after: `npm run check --workspace=packages/website` went from failing with "Cannot find native binding" to 0 errors/0 warnings/0 hints, and `npm run build` completes cleanly end to end (7/7 tasks). 130 tests across 5 new/extended suites, all passing.

**Found along the way:** two real bugs not anticipated in the original plan, both fixed in this PR — (1) a named Docker volume mounted under the read-only host bind needs its mountpoint directory to pre-exist on host or container creation fails outright; (2) the existing SMI-6034 troubleshooting entry's ACL-remediation command was left silently incomplete by this change (it now covers ~120 additional paths) and has been corrected to derive the list dynamically instead of drifting again on the next dependency bump.

**Net result:** Fully shipped. A completely fresh `./scripts/create-worktree.sh` dress rehearsal (beyond what this PR's own diff can verify pre-merge, since that script is main-checkout-rooted) is recommended as an immediate post-merge sanity check.

---

## Technical detail

**New**: `scripts/lib/linux-optional-packages.mjs` — derives the set of Linux-only optional-platform packages directly from `package-lock.json`'s `os` field (not a hand-maintained list), including nested-under-a-dependency positions (e.g. Astro's own vendored Rolldown/Lightning CSS copies). Both the Docker image build and host-side compose generation read from this single source of truth.

**Changed**:
- `Dockerfile` — seeds every derived package into `/opt/native-seed/tier-b/<real hierarchy>` at image-build time, plus a version marker, with existence/non-empty validation
- `docker-entrypoint.sh` — version-marker-aware restore loop into the (now-real) named volumes at container boot, re-seeding on a version mismatch
- `scripts/_lib.sh` — `enumerate_native_module_volumes()`/`enumerate_compose_node_modules_mounts()` extended to emit real named-volume declarations for every derived path (gated on the derivation script's output, never on host directory existence — the existing host-existence gate would silently emit nothing on macOS, since npm never creates these directories there at all); new `ensure_tier_b_mount_sources()` pre-creates mountpoint placeholders
- `scripts/create-worktree.sh` — wires in the new mount-source step
- `CLAUDE.md` — new Troubleshooting entry + corrected the SMI-6034 entry's now-widened ACL remediation
- `docs/internal/process/guards-and-opt-outs.md` — registers `SKILLSMITH_TIER_B_SEED_DISABLE`, a one-flag rollback

All four waves shipped with per-wave governance code review (see `docs/internal/code_review/2026-08-16-smi6050-wave{1,2,3}-*.md`). Rollback: `SKILLSMITH_TIER_B_SEED_DISABLE=1` reverts to pre-fix behavior with zero Tier-B masking, no revert required.

Refs SMI-6050


wrsmith108:
**Pre-push bypass note**: this branch was pushed with \`--no-verify\` on this specific push, with explicit user authorization, after the local pre-push test suite repeatedly hit genuine host-level resource exhaustion (\`Resource temporarily unavailable\` on the shell's own \`echo\` writes mid-test-output, not a test failure — confirmed via a live-monitored, fully clean pre-push pass earlier in the session, direct re-verification of every affected native module/mount after each host hiccup, and concurrent container count climbing from 13 to 17 across retries on this shared dev machine).

The code itself has been verified correct multiple times independently of this bypass:
- Full 130-test suite (5 new/extended files) passing
- The actual production bug reproduced before the fix and confirmed fixed after (\`npm run check --workspace=packages/website\`: 0 errors/warnings/hints)
- \`npm run build\`: 7/7 tasks successful
- A complete, clean pre-push run (security tests + format check + full test suite) did pass once, before an unrelated docs/internal submodule rebase

CI (this repo's own clean, non-contended runners) is the authoritative verification going forward — watching it now before this is ready for review.

PR #2395, opened 2026-08-17
chore(docs): bump docs/internal pointer for SMI-5334 date fix

wrsmith108:
## Business Summary

**What shipped**: Bumps `docs/internal` to pull in a one-line date correction (smith-horn/skillsmith-docs#774) — PR #2393's runbook addition said the B1 cross-prefix handoff was "discovered 2026-08-14" but the actual driver-log evidence is timestamped 2026-08-16. Pointer-only, no source changes.

**Quality bar**: Docs-only; full pre-push suite ran clean (security, format, tests).

**Net result**: Runbook date is now accurate.

[skip-impl-check]

PR #2403, opened 2026-08-17
chore(docs): bump docs/internal pointer for SMI-5334 date fix (retry)

wrsmith108:
Supersedes closed #2395. docs/internal's history diverged (a criss-cross merge incorporated the pre-date-fix commit without the fix itself); the correct date text is already present on the current tip via a different lineage. This bumps the outer pointer to that current tip. [skip-impl-check]


wrsmith108:
Redundant -- main's docs/internal pointer already advanced past this PR's target (confirmed via git merge-base --is-ancestor) and the corrected date text is already live on main. No further action needed for the SMI-5334 date fix.

PR #2416, opened 2026-08-17
fix(SMI-6050): address 3 findings from GPT-5.6-Sol post-merge review

wrsmith108:
## Business Summary

**What shipped:** Three real correctness fixes to the SMI-6050 mechanism (PR #2396), found by an independent GPT-5.6-Sol post-merge review — a genuine second opinion, not a rubber stamp. One blocking: a corrupted/partial image build could have its empty seed directory silently "blessed" as a successful restore, since copying zero files isn't an error. Two lower-severity: a version bump's restore never removed files the old version had that the new one dropped, and a derivation-script failure anywhere in `scripts/_lib.sh` would have silently produced zero Tier-B mounts instead of a loud, diagnosable error — exactly the class of silent-gap failure SMI-6050 exists to eliminate, just via a different root cause.

**Quality bar:** Every finding independently re-verified against the actual merged code before fixing (not taken on the reviewer's word). 6 new/extended tests, all exercising real file-content mutation and real shell execution, not mocks. 74/74 shell tests + 64/64 vitest tests pass. Lint and typecheck clean.

**Net result:** Fully shipped, no further follow-up expected from this review pass.

---

## Technical detail

- `docker-entrypoint.sh` — the Tier-B restore loop now treats an empty seed directory exactly like a missing one (skip, count as failed, no marker written); clears the target's contents (not the directory itself, which may be an active named-volume mountpoint) before every restore so a version bump's removed files don't survive
- `scripts/_lib.sh` — new shared `derive_tier_b_paths()` helper used by all three prior ad-hoc call sites; always surfaces a loud stderr error and non-zero return on a genuine derivation failure, never a silently-empty result
- `scripts/tests/docker-entrypoint-tier-b-seed.test.ts` — 2 new tests (empty-seed guard, stale-file-removal-on-version-bump)
- `scripts/tests/enumerate-compose-mounts-smi5650.test.sh` — 1 new test (test 26) covering the loud-failure behavior across all three call sites via a malformed-lockfile fixture

Refs SMI-6050

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -433,6 +433,19 @@ elif [ -f "/app/.git" ] && [ -d "/opt/native-seed/tier-b" ]; then
         target="/app/${rel_path}"
         target_marker="${target}/.smi6050-seed-version"
 
+        # Corrupted/partial seed guard (post-merge review finding): an empty
+        # seed directory would otherwise make `cp -a seed/. target/` succeed
+        # trivially (copying zero files is not an error) and then stamp the
+        # target "current" without restoring any real content — a silent
+        # false-success indistinguishable from a healthy restore. Treat an
+        # empty seed exactly like a missing one: skip and count as failed,
+        # never as restored or current.
+        if [ ! -d "$seed_dir" ] || [ -z "$(ls -A "$seed_dir" 2>/dev/null)" ]; then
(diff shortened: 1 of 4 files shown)


wrsmith108:
**Pre-push bypass note**: pushed with `--no-verify` after two consecutive host resource-exhaustion failures (`Resource temporarily unavailable` on the pre-push script's own `echo` writes, hitting unrelated packages `mcp-server`/root — this PR touches neither), with explicit user authorization. The actual diff (4 files) was independently re-verified fresh on a newly-recreated container immediately before this push: 74/74 shell tests + 6/6 vitest tests pass. CI's clean runners are the authoritative gate going forward.

PR #2465, opened 2026-08-22
fix(cli,edge): reject bare-name registry match without author confirmation

wrsmith108:
## Business Summary

**What shipped:** `skillsmith update` and its underlying registry lookup (`skills-get`) no longer
substitute an unrelated same-named skill when the one you actually asked for doesn't exist. Both
the CLI's local-cache diff and the server-side registry lookup now confirm the author before
trusting a name match — previously either path would silently fall back to any skill sharing the
same bare name, from any author, and `update` would force-install that unrelated content over
your local file.

**Quality bar:** Implementation matches a confirmed real-world incident (two personal, unclaimed
skills were overwritten live during a demo run) — the new tests reproduce that exact shape.
Typecheck, lint, format, and `audit:standards` all clean; full CLI suite (1016 tests) and the
`skills-get` edge-function suite (17 tests) pass, including 2 new CLI regression tests and 1 new
edge-function regression test for this fix, plus 2 pre-existing edge-function tests updated to
reflect the corrected (safe) behavior they were previously asserting against.

**Net result:** Fully shipped, no follow-up. The two incident skills were already restored
manually before this fix — this closes the hole that caused it.

---

## Technical detail

- `packages/cli/src/commands/manage.update.ts` — `getSkillDiff()` now reads the installed skill's
  claimed author from its own `SKILL.md` front-matter (new `readClaimedAuthor()` helper) and only
  trusts a local-registry-cache bare-name match when it agrees with the cache row's author.
  Otherwise it falls through to the existing confidence-gated manifest/`SourceRecoveryService`
  path rather than silently substituting the wrong author's skill.
- `supabase/functions/skills-get/index.ts` — the fuzzy bare-name fallback (tier 3) now runs ONLY
  when the caller's `skillId` never specified an author to begin with (`!skillId.includes('/')`).
  An `author/name`-shaped id that misses both the direct-id and exact-author+name lookups now
  gets a definitive 404 instead of a substituted unrelated-author skill. `skillsmith update`
  trusts a 200 here as "this is the skill," so this closes the same hole server-side.
- `packages/cli/tests/manage.update.test.ts` — 2 new regression tests (no claimed author; claimed
  author that doesn't match the cache row) plus fixture updates across the existing suite to add
  matching `author` fields now that the check exists. Added `manifestKeyFor`/`SourceRecoveryService`
  mocks (this file's first tests to reach that fallback code path).
- `supabase/functions/skills-get/index.test.ts` — 1 new regression test asserting `.ilike` (the
  fuzzy stage) is never invoked for an `author/name`-shaped id; 2 existing tests updated from
  asserting the old (unsafe) fall-through behavior to the corrected one.

Fixes SMI-6103.


wrsmith108:
Cross-model review via GPT-5.6-Sol (NEEDLE dispatch, analysis-only) on the first commit surfaced a real correctness bug in the initial fix:

> `manage.update.ts` selected the first same-name row from the local cache before checking its author. If an unrelated author's same-named skill happened to sort before the installed skill's own correct-author row, the correct row was never considered — a legitimate update would wrongly report `unresolvable`.

Fixed in a32c148f2: the lookup now scans every same-name row for the one whose author matches the installed skill's own claim, instead of stopping at the first name match. Added a regression test (`finds the correct-author row even when an unrelated-author row of the same name sorts first`) reproducing the exact collision shape.

Sol's review otherwise confirmed the security boundary holds: author-qualified `skills-get` requests miss to a 404 without fuzzy substitution, and genuine bare-name requests retain the existing fuzzy fallback.

wrsmith108:
## PR Review (pre-merge gate) — pr-reviewer

**Verdict: PASS-WITH-FIXES**

Ran the 14-check pre-merge gate against merge-base `fa22a9ec38282ec8e0ecd715ab05a9a22168daf0`. 1 passed (PR-07), 1 fixed (PR-14), 12 skipped (no SQL/enum/edge-fn-surface/rename triggers in this diff; PR-12 skipped — no implementation plan, this is a live-incident hotfix and SMI-6103 has full root-cause documentation instead).

**PR-07 (Silent catch)**: The one new catch block (`readClaimedAuthor()` in `manage.update.ts`) is a read-only optional-`SKILL.md`-metadata fallback, not a write-path swallow — a read failure returns `null`, which forces the caller to skip the bare-name cache shortcut and fall through to the existing confidence-gated recovery path. Fails safe, not silently unsafe.

**PR-14 (audit:standards) — FAIL→FIXED**: The worktree's own shallow-clone `audit:standards` run (0 fail / 9 warnings, all pre-existing/environmental) couldn't reproduce it, but this PR's CI **Quality Checks** job (full git history) caught Check 54 (CHANGELOG Entry Gate, SMI-5680): `packages/cli/src/**` changed with no new `[Unreleased]` entry in `packages/cli/CHANGELOG.md`. Fixed in commit `2b2ae6e9a`.

Also independently re-ran the full touched-suite tests (48 passing across `manage.update.test.ts`, `manage.update.source-recovery.test.ts`, and `supabase/functions/skills-get`), typecheck, lint, and format:check — all clean.

Full report: [skillsmith-docs#806](https://github.com/smith-horn/skillsmith-docs/pull/806) (`docs/internal/pr-reviews/2026-08-22-smi-6103.md`)

wrsmith108:
## PR Review update — CI now fully green

All checks now pass (39/39 completed, 2 intentionally skipped/skipping: Website Skills E2E Run, Fresh Install Test). \`mergeStateStatus: CLEAN\`, \`mergeable: MERGEABLE\`, zero failures.

Combined with the PASS-WITH-FIXES pre-merge gate verdict above (commit \`2b2ae6e9a\` resolved the one real finding), **this PR is ready to merge.** Merge is a human/coordinator decision — not performed by this review.

PR #2466, opened 2026-08-22
fix(db): sync profiles.email_verified on real email confirmation

wrsmith108:
## Business Summary

**What shipped:** New email/password sign-ups who confirm their address through the normal
confirmation-link flow now get their account correctly marked as verified — previously that flag
stayed permanently stuck "unverified" for this entire signup path, with no visible error pointing
at the cause, silently blocking license-key issuance forever.

**Quality bar:** Reviewed via the `supabase-migration-reviewer` skill (APPROVE-WITH-NOTES — low
risk given `auth.users`' confirmed low write rate). Applied to staging and verified end-to-end
with a disposable test account: create unconfirmed → confirm → the verified flag flips correctly,
matching the real confirmation timestamp. Staging's backfill closed the gap for every
already-affected row (confirmed via a live count query: 0 stale rows remaining).

**Net result:** Fully shipped to staging. Prod apply is intentionally deferred — this repo's
migration-safety playbook requires a 24h staging soak before any `auth.users`-touching migration
reaches prod. Not urgent: the one account that was blocking a live demo was already unblocked
manually days ago; this PR is the real, permanent fix plus backfill for the remaining affected
accounts.

---

## Technical detail

- `supabase/migrations/20260822000000_sync_email_confirmation.sql` — new `SECURITY DEFINER`
  trigger `on_auth_user_email_confirmed` on `auth.users`, firing only on the genuine
  `email_confirmed_at` NULL → non-null transition (disjoint from the two existing `auth.users`
  triggers — `on_auth_user_updated`/`sync_github_profile` gates on `raw_user_meta_data`,
  `on_auth_user_email_changed` gates on `email` itself changing — so all three coexist without
  conflict). Sets `profiles.email_verified = TRUE` / `email_verified_at = NEW.email_confirmed_at`,
  guarded idempotent (`WHERE email_verified IS NOT TRUE`).
- Same migration includes a one-time backfill `UPDATE` closing the gap for every row already
  affected (bounded by its own join + `IS [NOT] NULL` predicates — never touches a legitimately
  unconfirmed account).
- `REVOKE ALL ... FROM PUBLIC` / `REVOKE EXECUTE ... FROM anon, authenticated` on the new function
  (audit:standards Check 52 / SMI-5526 — no direct-call surface needed, only trigger-internal
  invocation).
- Root cause: `handle_new_user()` (`080_profile_completion.sql`) only sets `email_verified = TRUE`
  for GitHub OAuth at signup; every other provider gets `FALSE` (correct at that instant), and
  the only other write path (`sync_profile_email()`) fires on a self-service email *change*, not
  the initial confirmation. Confirmed against live prod data: 37 of 42 email/password profiles
  affected.
- Live-DB pre-apply checks (via `pooler-psql.sh`): trigger/function name uniqueness confirmed, no
  collision; `auth.users` write rate confirmed low (91 updates / 6 inserts lifetime) — well within
  the 3s `lock_timeout` guard.
- Rollback block included; does not revert already-backfilled rows (reverting would reintroduce
  the bug for accounts that already confirmed).

Fixes SMI-6102.

[skip-impl-check]

This PR is a SQL-only Supabase migration (new trigger + function + one-time backfill on `auth.users`/`profiles`) — the migration file itself IS the implementation; there is no accompanying TypeScript/JS source change for `verify-implementation.ts`'s file-extension heuristic to detect. The migration was applied to staging and verified end-to-end (see PR description) before this marker was added.


wrsmith108:
Cross-model review attempt via GPT-5.6-Sol (NEEDLE dispatch) was killed mid-turn with no usable output (`success-without-agent-message` — a transient Codex-side failure, not a review verdict). Per this repo's dispatch policy, that outcome isn't final and falls back to normal Claude-tier review, done directly here:

**Trigger disjointness (verified against all 3 `auth.users` triggers + the `sync-oauth-email` admin-API email-flip path):**
- `on_auth_user_created` (AFTER INSERT) — different event entirely, no interaction.
- `on_auth_user_updated`/`sync_github_profile` (gates on `raw_user_meta_data`) — different column, no overlapping writes even if both fire on the same statement (this trigger never touches `email_verified`/`email_verified_at`).
- `on_auth_user_email_changed`/`sync_profile_email` (gates on `email` itself changing) — the interesting case. Traced `sync-oauth-email/index.ts`'s `updateUserById(..., { email, email_confirm: true })` call, which can touch both `email` and `email_confirmed_at` in one statement. If a user's `email_confirmed_at` was genuinely NULL before that flip (first-time confirmation via this path), BOTH triggers fire on the same statement — harmless: both are idempotent writes to the same two columns (`email_verified=TRUE`, `email_verified_at=~now`), execution order is deterministic (alphabetical: `email_changed` before `email_confirmed`), and the net state is identical to the already-established "disjoint WHEN, no recursion" pattern this repo already uses for the sibling `sync_profile_email()` migration. For the common case (a previously-OAuth-confirmed user), `OLD.email_confirmed_at` is already non-null, so the new trigger's WHEN clause correctly stays silent.
- `double_confirm_changes = true` (config.toml:201): an email-CHANGE flow only reaches this trigger's WHEN clause if `email_confirmed_at` was genuinely NULL beforehand — for an already-confirmed user changing addresses, OLD is non-null, so no re-fire. No conflict with `sync_profile_email()`'s own handling of the new address.

**Backfill WHERE clause**: correctly scoped — `u.email_confirmed_at IS NOT NULL AND p.email_verified IS NOT TRUE` touches exactly the affected population, never a legitimately-unconfirmed row, never an already-verified row (avoids redundant `updated_at` bumps).

**One informational-only edge case, not a fix in scope for this PR**: an account created via the Admin API with `email_confirm: true` set at INSERT time (never followed by a separate UPDATE) would still get `email_verified=FALSE` from `handle_new_user()`'s provider-only branch, since no UPDATE ever fires this new trigger. Confirmed via grep this path doesn't exist anywhere in this codebase's real signup flows (the only `email_confirm:true` caller, `sync-oauth-email`, always does it via `updateUserById`, not `createUser`) — so this isn't a live gap, just worth naming for anyone adding an admin-driven instant-confirm signup path in the future.

No blocking issues found. Verdict: safe to proceed through the standard 24h staging soak → prod apply.

PR #2494, opened 2026-08-23
fix(mcp-server,db): remove SUPABASE_SERVICE_ROLE_KEY from registry install path

wrsmith108:
## Business Summary

**What shipped:** The private skill registry's `install` action (both the CLI and the MCP tool) no longer needs the most powerful backend credential (`SUPABASE_SERVICE_ROLE_KEY`) configured anywhere on a customer's machine — the last remaining piece of a security hardening pass that started with SMI-6109. A new, narrowly-scoped database function answers exactly one question server-side ("does this specific team's subscription justify Enterprise-tier content access") without needing that credential at all.

**Quality bar:** VP-reviewed implementation plan (Product/Engineering/Design) before any code was written — Engineering independently re-derived the plan's central security claim from raw SQL rather than trusting the explanation, and confirmed it. All 1017 mcp-server tests pass, website build is clean, `audit:standards` passes (a real gap it caught — a missing explicit anon-access revoke on the new database function — was fixed during implementation, not deferred). Found and fixed two stale documentation claims along the way: the docs page still described an already-fixed part of the registry as needing the old credential, and an architecture decision record cited an addendum that was promised in an earlier PR but never actually written.

**Found along the way:** none deferred — every inaccuracy found during implementation (a missing security grant, two stale docs pages, one missing architecture-doc addendum) was fixed in this same PR rather than filed separately.

**Net result:** Fully shipped pending merge + publish. This lives inside `@skillsmith/mcp-server`, which customers install via `npx`, so a version bump and publish run (tracked in the plan's Verification checklist) is required after merge before this reaches real users — not optional, and time-sensitive this week for planned design-partner UAT.

---

## Technical detail

- `supabase/migrations/20260824000000_check_registry_team_entitlement.sql` (new) — `check_registry_team_entitlement(p_team_id TEXT)`, a `SECURITY DEFINER` function granted to `authenticated` only (explicit `REVOKE ... FROM anon`, not just the implicit `PUBLIC` revoke — caught by `audit:standards` Check 52 during implementation). Deliberately not built on the existing `resolve_effective_entitlement`, whose personal-subscription fallback would leak entitlement across teams for this specific use case — full rationale in the plan doc.
- `packages/mcp-server/src/tools/registry-tools.live.content.ts` — `isTeamEnterpriseEntitled()` now calls the RPC via the caller's own member-authenticated client instead of creating a service-role client. Removed the now-dead admin-client helper and constants.
- `packages/mcp-server/src/tools/registry-tools.live.content.test.ts` — entitlement-path mocks rewritten around the RPC contract; added a regression assertion that the RPC is only ever reached via the member client.
- `packages/mcp-server/src/tools/registry-tools.live.reads.ts` — corrected a stale header comment still describing the pre-SMI-6109 service-role architecture for `list`/`get`.
- `scripts/audit-standards.mjs` — removed the now-unnecessary allowlist entry for this file.
- `packages/website/src/pages/docs/private-registry.astro` + `tutorials/private-registry.astro` — corrected two stale claims: list/get/namespace were still described as service-role-based (already fixed by SMI-6109, this page hadn't caught up), and the MCP install path was described as unreliable (no longer true).
- Full plan, VP review findings, and rejected-alternative rationale: `docs/internal/implementation/smi-6111-registry-content-install-entitlement-rpc.md`. ADR-116 addendum (also fulfilling an SMI-6109 addendum that was promised but never written): `docs/internal` PR https://github.com/smith-horn/skillsmith-docs/pull/841 (separate repo, left unmerged for the coordinator).

Refs SMI-6111


wrsmith108:
## PR Review — pr-reviewer skill

**Verdict: PASS-WITH-FIXES** — ready to merge (not merging myself, that's a human decision).

Full migration checklist (PR-01/02/03/04/05/06/13) independently re-verified against merge-base and prior migration history — migration ordering, `search_path`, `schema_version` sequencing (101, correctly following merge-base max 100), lock timeouts, and both SQL-comment object references (`resolve_effective_entitlement`, `teams.subscription_id ... ON DELETE SET NULL`) all check out.

**Central security claim independently re-derived from raw SQL, not just the plan**: read `resolve_effective_entitlement`'s migration and `tier_rank()` directly. Confirmed the personal-subscription leg of that function's UNION has no team correlation and isn't filtered by `p_team_id`, and `tier_rank(enterprise)=3` is the max rank — so a caller with a personal Enterprise subscription really would resolve to `entitled=true` for an unrelated non-Enterprise team via that function's personal leg. The new `check_registry_team_entitlement` RPC correctly avoids this.

**Plan-vs-diff (PR-12)**: all of the plan's substantive directives (sections 1-6) are covered. One plan file-list item — `registry-tools.live.adversarial-content.test.ts` — wasn't touched; independently confirmed this is a plan overreach, not a real gap (that file never exercises the entitlement path; its 17 tests pass unmodified). Also independently verified the PR description's claim that `registry-tools.ts`'s class comment didn't need changing — read it directly, confirmed already accurate.

**One real finding, fixed**: `audit:standards` Check 54 (CHANGELOG Entry Gate) was failing in CI's `Quality Checks` job — no `[Unreleased]` entry had been added to `packages/mcp-server/CHANGELOG.md`. This wasn't visible in a local worktree-container `audit:standards` run (the check self-skips without a resolvable `origin/main` ref in that context), but showed up on live CI. Fixed in `1da5725f5`: added the missing entry and corrected an adjacent stale claim in the existing SMI-6109 entry ("until SMI-6111 lands" — this PR is that landing). All 46 CI checks are now green.

Full report: `docs/internal/pr-reviews/2026-08-23-smi-6111-registry-entitlement-rpc.md` (docs-internal PR [smith-horn/skillsmith-docs#842](https://github.com/smith-horn/skillsmith-docs/pull/842), left unmerged for the coordinator, same as this PR's own ADR-116 addendum PR #841).

Issue #861, opened 2026-05-01
perf(ci): docker-build job wastes ~4-5min on every cache-hit run


peter604:
## Summary

The `docker-build` job in `ci.yml` has four compounding inefficiencies that together waste **~4–5 minutes on every cache-hit run** (the common case). Identified by profiling step timings against the cache-hit path.

Linear: SMI-XXXX

---

### 1. Unconditional re-save of Docker image artifact — **2m 34s** (biggest offender)

**Step:** `Save Docker image for artifact` (line ~742)

The step has no `if:` guard. On a cache hit, the image is loaded from `/tmp/docker-cache/skillsmith-ci.tar.gz` and then immediately re-serialized via `docker save | gzip` into `/tmp/skillsmith-ci.tar.gz`. This is a full compress/decompress round-trip for a file that already exists on disk.

**Fix:** Copy the cached tarball directly on cache hit:
```yaml
- name: Save Docker image for artifact
  run: |
    if [ "${{ steps.docker-image-cache.outputs.cache-hit }}" = "true" ] && \
       [ -f /tmp/docker-cache/skillsmith-ci.tar.gz ]; then
      cp /tmp/docker-cache/skillsmith-ci.tar.gz /tmp/skillsmith-ci.tar.gz
      echo "✓ Copied from cache (no re-save needed)"
    else
      docker save skillsmith-ci:${{ github.sha }} | gzip > /tmp/skillsmith-ci.tar.gz
    fi
```
**Est. savings: ~2m 25s per cache-hit run.**

---

### 2. gzip for Docker image cache — slow decompress on load — **2m 0s**

**Step:** `Load cached Docker image`

`gunzip -c skillsmith-ci.tar.gz | docker load` is CPU-bound. gzip is single-threaded and slow on large images. zstd decompresses 3–5× faster at comparable ratios.

**Fix:** Switch cache storage to zstd in both save and load paths:
```bash
# Save (after build)
docker save skillsmith-ci:$SHA | zstd -T0 -3 > /tmp/docker-cache/skillsmith-ci.tar.zst

# Load (cache hit)
zstd -d /tmp/docker-cache/skillsmith-ci.tar.zst --stdout | docker load
```
Update the cache `key` path from `.tar.gz` → `.tar.zst`. Requires `apt-get install -y zstd` (already available on `ubuntu-latest`).

**Est. savings: ~60–80s per cache-hit run.**

---

### 3. node_modules extraction re-compresses with gzip — **2m 10s**

[...]

PR #2540, opened 2026-08-26
fix(website): sitewide indexing bugs from Search Console audit (SMI-6180)

wrsmith108:
## Business Summary

**What shipped:** Three fixes for why Google wasn't indexing large parts of skillsmith.app: every page's canonical tag now matches the URL it's actually served at (was pointing at a duplicate URL sitewide); individual skill pages now tell Google their real title, description, and content up front instead of a generic loading screen; and two broken redirect links (`/changelog`, `/docs/authoring`) now work for both URL forms Google was crawling.

**Quality bar:** Verified against the live generated HTML before and after each fix, not just "tests pass" — confirmed the canonical tag actually changed and duplicate tags are gone. A real bug (skill URLs with a slash in the ID breaking due to double-encoding) was caught by writing tests before shipping, not after. Full typecheck, lint, format check, and both test suites (690 + 17,481 tests) pass. A governance review after the first commit found and fixed a second issue: a way malicious content in a skill's name could break out of its structured-data tag — fixed here, with the same pattern elsewhere in the site filed as a separate follow-up.

**Found along the way:** Two issues filed separately rather than bundled in here — SMI-6188 (a pre-existing, repo-wide bug that currently blocks the website's production build inside Docker, unrelated to this change) and SMI-6189 (the same script-injection pattern this PR fixes exists in other places across the site and needs its own pass).

**Net result:** All three confirmed bugs from the audit are fixed. The two follow-ups are tracked, not blocking.

---

## Technical detail

**Files changed:**
- `packages/website/astro.config.mjs` — adds `trailingSlash: 'never'`. Root cause: with it left at Astro's default, every prerendered page's self-referencing canonical tag carried a trailing slash the actual served URL never had. Confirmed live via curl before/after; `build.format: 'file'` was tried first and does *not* fix this under `@astrojs/vercel` (the adapter overrides it) — see the in-file comment for the full trace.
- `packages/website/vercel.json` + root `vercel.json` — `"trailingSlash": false` (stops Vercel itself serving both URL forms with no redirect), plus explicit `/changelog/` and `/docs/authoring/` redirect rules (the existing rules only matched the no-slash form, so Google's crawled slash-variants 404'd).
- `packages/website/src/pages/skills/[id].astro` — individual skill pages fetch metadata server-side now (mirrors the existing category-page pattern in the same file) instead of rendering a static "Skill Details" shell for every skill. Sets real title/description/canonical/JSON-LD, and a genuine 404 (not a 200 that shows an error client-side) for skills that don't exist. Also removes a second, conflicting canonical tag that was being injected on category pages alongside `BaseLayout`'s own correct one.
- `packages/website/src/pages/status.astro` — trailing slash removed from its explicit canonical override, matching the sitewide fix.
- `packages/website/src/lib/skill-page-meta.ts` (new) + `skill-page-meta.test.ts` (new, 14 tests) — the URL-decode/encode, title/description/canonical/JSON-LD derivation, and JSON-LD escaping logic pulled out of the page frontmatter so it's directly unit-tested. Includes `safeJsonLdScript()`, the fix for the script-injection issue.

**Verification:** `astro check` (0 errors/warnings across 206 files), `tsc --build`, `eslint .`, `prettier --check .`, `vitest run` in `packages/website` (690 tests) and at the repo root (17,481 tests) — all clean. Generated HTML inspected directly (both via a full `astro build` for prerendered pages and a local dev server for the SSR skill/category routes) to confirm canonical tags, title tags, and 404 status codes match expectations.

**Not run:** `npm run build` / `npm run preflight` for `packages/website` — currently fails in Docker on an unrelated, pre-existing bug (SMI-6188), confirmed via `git stash` to reproduce identically without this PR's changes, on both the main checkout's container and a fresh worktree's. Substituted the targeted commands listed above.

Linear: SMI-6180 (parent), SMI-6188 and SMI-6189 (follow-ups, out of scope here).

### DIFF
--- a/packages/website/astro.config.mjs
+++ b/packages/website/astro.config.mjs
@@ -81,6 +81,30 @@ export default defineConfig({
   // Build output configuration - static with SSR adapter for dynamic routes
   output: 'static',
 
+  // SMI-6180: with trailingSlash left at its default ('ignore'), Astro's own
+  // default build.format ('directory') makes Astro.url.pathname carry a
+  // trailing slash for every prerendered page during the static build —
+  // BaseLayout's self-referencing canonical (src/layouts/BaseLayout.astro)
+  // inherits that slash, while Vercel serves every page WITHOUT one. Every
+  // page's own canonical tag ends up pointing at a trailing-slash duplicate
+  // of itself, confirmed live against production (curl'd multiple pages: 200
+  // at the no-slash URL, canonical pointing to .../<path>/, which also
+  // resolves 200).
+  //
+  // build.format: 'file' alone does NOT fix this under @astrojs/vercel —
+  // confirmed by building both ways: the adapter's own static-output
+  // structure (still page/index.html either way) overrides build.format's
+  // effect on Astro.url.pathname for these pages. trailingSlash: 'never' is
+  // unconditional in Astro's own core build-format helper (see
+  // astro/dist/core/build/util.js's slash-append decision) regardless of
+  // buildFormat — confirmed this is what actually removes the trailing
+  // slash from the rendered canonical tag.
+  // vercel.json's own top-level "trailingSlash": false (both copies) is the
+  // complementary fix: it stops Vercel itself from serving a page at both
+  // the slash and no-slash URL with no redirect between them, which is the
+  // other half of this bug (verified live: both forms returned 200).
+  trailingSlash: 'never',
+
   // SMI-6110: account dashboard UX consolidation (H2) — the `/account/team`
   // → `/account` true HTTP 301 is declared in vercel.json (both the root and
   // packages/website copies, per SMI-4641 structural sync), NOT here.
--- a/packages/website/src/lib/skill-page-meta.test.ts
+++ b/packages/website/src/lib/skill-page-meta.test.ts
@@ -5,6 +5,7 @@ import {
   deriveSkillPageMeta,
   resolveSkillId,
   safeJsonLdScript,
+  shouldCacheSkillPage,
   type SkillMeta,
 } from './skill-page-meta'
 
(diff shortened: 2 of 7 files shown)


wrsmith108:
## Pre-merge review (GPT-5.6-Sol via NEEDLE)

Full report: `docs/internal/pr-reviews/2026-08-26-smi-6180-indexing-fixes.md` (up for review: smith-horn/skillsmith-docs#861)

**14-check gate**: 4 PASS, 9 SKIP (no SQL/migrations/edge functions/enums in this diff), `audit:standards` re-run directly with Docker access — 0 failures, 7 pre-existing warnings unrelated.

**Beyond the checklist**, an adversarial pass found 4 real issues, 3 fixed here (commit d284a9264), 1 filed separately as architectural:

- **High** — the SSR skill-metadata fetch authenticates with the public anon key, which shares **one global** 30 req/min rate bucket across all anonymous traffic sitewide (not per-caller). A crawler burst on uncached skill pages could 429 this fetch and silently fall back to the pre-fix generic shell — undermining this PR's own purpose under the exact conditions it matters most. Partially mitigated by the cache fix below; the full fix (a dedicated server-side identity/quota) needs its own design pass — filed as SMI-6190.
- **Medium**, fixed — degraded responses (network error, 5xx, rate-limited) were being CDN-cached for 5+ minutes, baking a transient failure into the exact generic-shell problem this PR fixes. Added `shouldCacheSkillPage()`.
- **Medium**, fixed — a malformed percent-encoded skill ID (`%E0%A4%A`, a lone `%`) threw an uncaught `URIError` outside the frontmatter's `try/catch`, crashing the SSR render. Now falls back gracefully to a real 404.
- **Low** — `audit:standards`' File Length check never scans `.astro` files at all, which is why this file's already-1098-line size (984 before this PR) went uncaught. Filed as SMI-6191.

JSON-LD escaping, 404/`noindex` handling, and the trailing-slash config itself were all independently confirmed correct. 6 new unit tests cover the 3 applied fixes.

PR #2658, opened 2026-08-31
fix(scripts): docs/internal ancestry-check engine + pre-push tier fix (SMI-6260 Wave 1)

wrsmith108:
## Business Summary

**What shipped:** A new automated check that catches a real, recurring bug: `docs/internal` (this repo's internal-docs submodule) can silently regress backward when a PR merges with a stale snapshot, dropping already-merged content with no visible trace in the PR diff. This first PR lays the groundwork — a shared, well-tested ancestry-checking engine, plus a fix so pushing a docs-only change no longer runs the full ~3-5 minute test suite for what should be a ~30 second push. The engine is wired up as an advisory (warn-only) local check in this PR; the hard, PR-blocking CI gate follows in a second PR.

**Quality bar:** Implementation by a Sonnet subagent, independently re-verified line-by-line against the plan's binding design (not trusted from the subagent's report alone) — re-ran the full 22-test suite myself, re-ran `audit:standards`/`preflight`/`typecheck`/`lint`/`format:check`, and read the core rule engine in full. The plan itself went through two adversarial review passes before implementation began (a same-family Claude fallback review, since the normal cross-model GPT-5.6-Sol dispatch failed for tooling reasons — see SMI-6333), catching three real logic gaps in the original rule design before any code was written. A full governance code-review pass found zero findings; report at `docs/internal/code_review/2026-08-30-smi-6260-wave1-pre-push-and-ancestry-engine.md`.

**Found along the way:** an unrelated pre-existing quirk in `docs-only.yml`'s path filter (harmless today, filed as [SMI-6328](https://linear.app/smith-horn-group/issue/SMI-6328)) and a tooling gap in the cross-model review dispatch process itself ([SMI-6333](https://linear.app/smith-horn-group/issue/SMI-6333)) — neither bundled into this PR.

**Net result:** This PR alone is a pure improvement (faster docs pushes, an advisory heads-up on a stale pointer) with no enforcement teeth yet. The actual gate that blocks a bad merge lands in the follow-up PR, tracked on the same Linear issue (SMI-6260).

---

## Technical detail

**New files:**
- `scripts/lib/docs-only-patterns.sh` — canonical docs-only classification, now shared by all three bash callers instead of three independently-drifting copies. Tightened to exclude `packages/**/*.md` and `**/tests/**`/`**/fixtures/**` from the docs-only bucket (closes an SMI-4961-class hazard: a new `.md` at a gated fixture path would otherwise be misclassified as safe-to-skip).
- `scripts/ci/check-submodule-pointer.sh` + `scripts/ci/check-submodule-pointer.helpers.sh` — the ancestry-check engine. Implements an R0-R11 + R-FETCH rule table with a two-layer evaluation model (preconditions gate comparisons; comparisons are independently AND-combined, not first-match) per the plan's binding design. Generic over every `.gitmodules` mount, but only `docs/internal` is blocking in v1 — the three `skillsmith-strategy` mounts are warn-only.
- `scripts/tests/check-submodule-pointer.test.ts` — 22 tests against real local bare-repo git fixtures (not mocked git output), including a combined-state case, precondition-gating cases, a simulated fetch-failure case, and a synthetic replay of the real PR #2609 regression this whole plan exists to catch.

**Modified:**
- `scripts/pre-push-check.sh`, `scripts/lib/hook-docker-detect.sh` — source the shared lib instead of inlining/duplicating the regex; the security-test-suite check (Phase 2 CHECK 1) now skips on a docs-only push.
- `scripts/pre-push-coverage-check.sh` — gains its own independent docs-only computation (it runs as a separate child process, so it can't inherit the other script's classification) and skips the full per-package coverage loop on a docs-only push.
- `.husky/pre-push` — new warn-only step: if a push touches any `.gitmodules` mount, runs the new engine in `--mode=warn` and prints the result, but never blocks the push.

**Override:** `SKILLSMITH_PRE_PUSH_FORCE_FULL=1` forces the full pre-push suite regardless of docs-only classification.

**Plan:** `docs/internal/implementation/smi-6260-docs-internal-pointer-regression-gate.md`
**Linear:** SMI-6260 (this is PR 1 of 2 — Wave 2 adds the required CI gate, post-merge auto-repair, and the bump script)

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -135,6 +135,64 @@ if [ $SECURITY_STATUS -ne 0 ]; then
   exit 1
 fi
 
+# =============================================================================
+# SMI-6260 Wave 1: warn-only docs/internal (and other .gitmodules mount)
+# pointer-ancestry check, between Phase 2 and Phase 3. Advisory only — NEVER
+# blocks the push (the hard gate is Wave 2's CI job, not built yet).
+#
+# Cannot reuse DOCS_ONLY/CHANGED_FILES from Phase 2's pre-push-check.sh —
+# it just ran via `bash "$(dirname "$0")/../scripts/pre-push-check.sh"`
+# above, a CHILD PROCESS, so nothing it computed is visible back in this
+# shell. Independently recomputes the changed-file list here, matching
+# scripts/lib/hook-docker-detect.sh's own existing duplication pattern
+# (each caller sources the shared lib and computes its own answer).
+# =============================================================================
+GITMODULES_FILE="$(dirname "$0")/../.gitmodules"
+if [ -f "$GITMODULES_FILE" ]; then
+  if PP_SUB_UPSTREAM=$(git rev-parse --abbrev-ref --symbolic-full-name '@{u}' 2>/dev/null); then
+    PP_SUB_CHANGED=$(git diff --name-only "$PP_SUB_UPSTREAM..HEAD" 2>/dev/null || true)
+  else
+    git fetch origin main --quiet 2>/dev/null || true
+    PP_SUB_CHANGED=$(git diff --name-only origin/main..HEAD 2>/dev/null || true)
+  fi
+
+  # Generic over .gitmodules entries — exact-match discipline mirroring
+  # scripts/ci/classify-changes.ts's getSubmoduleMounts()/isSubmoduleMount()
+  # (never glob-match a mount path; it's data read from .gitmodules, not an
+  # authored pattern).
+  PP_SUB_MOUNTS=$(grep -E '^[[:space:]]*path[[:space:]]*=' "$GITMODULES_FILE" \
+    | sed -E 's/^[[:space:]]*path[[:space:]]*=[[:space:]]*//' \
+    | sed -E 's/^"(.*)"$/\1/')
+
+  PP_SUB_TOUCHED=0
+  if [ -n "$PP_SUB_CHANGED" ] && [ -n "$PP_SUB_MOUNTS" ]; then
+    while IFS= read -r pp_mount; do
+      [ -z "$pp_mount" ] && continue
+      if printf '%s\n' "$PP_SUB_CHANGED" | grep -qxF "$pp_mount"; then
+        PP_SUB_TOUCHED=1
+        break
+      fi
+    done <<PP_SUB_MOUNTS_EOF
+$PP_SUB_MOUNTS
+PP_SUB_MOUNTS_EOF
+  fi
+
+  if [ "$PP_SUB_TOUCHED" = "1" ]; then
+    echo "🔗 Checking submodule pointer ancestry (warn-only)..."
+    POINTER_CHECK_SCRIPT="$(dirname "$0")/../scripts/ci/check-submodule-pointer.sh"
+    if [ -r "$POINTER_CHECK_SCRIPT" ]; then
+      git fetch origin main --quiet 2>/dev/null || true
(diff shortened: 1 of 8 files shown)


wrsmith108:
## PR Reviewer — pre-merge gate (same-family Claude fallback, GPT-5.6-Sol dispatch blocked, SMI-6333)

**Checks**: 3 substantively executed (PR-07 PASS, PR-12 FAIL→FIXED, PR-14 PASS), 11 SKIP (no SQL/edge-fn/enum/module-move triggers present), 0 findings pending human review.

**Verdict: PASS-WITH-FIXES**

### Fixed (zero-deferral)

**PR-12 (plan vs diff)**: found a real deviation between the plan's binding R0-R11 design (each rule is one verdict per mount) and the actual behavior — `evaluate_layer2()` in `check-submodule-pointer.helpers.sh` printed *both* R4's own `PASS-WARN` line and a second, redundant generic `PASS` line for the same mount whenever S was ahead of T on a live branch with no independent B-axis (R7) failure. Confirmed live with a minimal repro fixture before fixing (none of the 22 existing tests caught it — they only assert `toContain`, not exact/singular output).

Fixed in `7f0bb7cee`: tracks whether R4 already printed its verdict and suppresses the trailing generic PASS only in that case — the B-axis (R7) check still runs unconditionally either way, so the combined-state (T<S<B) anti-regression case the plan explicitly required coverage for is unaffected. Added a regression assertion (exactly one result line per mount) to the existing R4 test. All 22 tests pass; `audit:standards` unchanged (80/8/0, 0 new fail/warn) before and after.

### Flagged for the coordinator (not fixed here — outside this review's scope)

1. **`mergeable: CONFLICTING`** — `origin/main`'s `docs/internal` gitlink has advanced past this PR branch's own pointer via concurrent docs-bump PRs. Ran the PR's own new `check-submodule-pointer.sh --mode=warn` against `origin/main` from this worktree: the PR branch's pointer is semantically fine (no R3/R7 hit against the submodule's own mainline) — this is a pure git-level gitlink 3-way-merge limitation, not a regression this PR introduced. Needs a rebase/gitlink bump before merge; that's a coordinator call, not a PR-content fix.
2. **Zero GitHub Actions check runs observed** for this PR as of review time (~30+ min post-open, ~25+ min post-fix-push), despite Actions being enabled repo-wide and firing normally for a sibling PR in the same window. No `[skip ci]` marker anywhere in the diff. This review's verdict rests on local-equivalent evidence instead (full local pre-push suite — security tests, format, per-package coverage across core/cli/mcp-server/enterprise — plus `audit:standards`, typecheck, lint all green in this worktree's own container). Recommend confirming real CI actually triggers before merge.

Full report: `docs/internal/pr-reviews/2026-08-31-smi-6260-wave1-docs-internal-pointer-gate.md` (pushed to the `docs/internal` submodule remote at `5fcef2a`; parent-repo gitlink bump left for the coordinator per this repo's convention).

wrsmith108:
**Non-blocking CI note**: the "Hooks Test" check failed (2 of 14 sub-tests), but this is not a required branch-protection check, and the failure is unrelated to this PR — it's the known, already-tracked \`mcp-command-guard\` test-fixture gap (SMI-6291: missing \`scripts/lib/mcp-command-guard.plugin-scan.mjs\` sibling module), which this PR never touches. Confirmed via \`git diff main...HEAD\` (zero overlap) and by checking recent unrelated \`main\`-branch CI runs, which show the same failure pattern independent of this branch. Not fixing here (out of scope, pre-existing, already tracked) — proceeding once the actual required checks are green.

PR #2659, opened 2026-08-31
feat(ci): docs/internal ancestry gate + post-merge auto-repair (SMI-6260 Wave 2)

wrsmith108:
## Business Summary

**What shipped:** The actual enforcement gate for SMI-6260: a required check that blocks any PR from merging a `docs/internal` gitlink bump that regresses backward relative to that submodule's own history — closing the exact bug that silently dropped merged content in a real prior incident (PR #2609). A second job re-checks every push to `main` after merge, since the PR-time check alone can be bypassed (admin merge, a direct push, or a timing gap), and either auto-repairs the common case or opens a GitHub issue with the exact fix command for anything it can't safely resolve on its own.

**Quality bar:** Implementation by a Sonnet subagent, independently re-verified line-by-line given the write-access risk — read the workflow YAML and the auto-repair script in full, traced every use of the CI's default credential versus the separate push-capable one, and walked the push logic for race safety under concurrent pushes. A full governance review then found and fixed two real issues before this ever reached a PR: an unused, overly broad permission grant on the job's default token, and a gap where a silent failure earlier in the repair sequence could let the job push (or claim to push) with stale content. Tested against the actual real-world incident — the true SHAs from PR #2609's history, not a stand-in.

**Found along the way:** confirmed live that the credential this job would eventually use to auto-repair `main` doesn't exist yet, so the job ships in alert-only mode today — it opens an issue instead of pushing, which is still a full step up from the current silence. A separate, unrelated infra bug in how this repo's worktrees resolve their pre-push checks was found and filed as its own issue ([SMI-6334](https://linear.app/smith-horn-group/issue/SMI-6334)), not folded into this PR.

**Net result:** SMI-6260 is fully shipped with this PR. The detection and alerting are live the moment this merges; full automatic repair activates once the missing credential is provisioned (documented in ADR-143, including exactly what that credential should and shouldn't be able to do).

---

## Technical detail

**New files:**
- `.github/workflows/submodule-pointer-check.yml` — two jobs. `pointer-check` (required, PR-time): fetches `docs/internal` via the existing `STRATEGY_SUBMODULE_PAT` pattern, runs `check-submodule-pointer.sh --mode=block`, and surfaces the failing rule's exact remediation as a `::error::` annotation on the PR. `pointer-autorepair` (`push: branches: [main]`): reconstructs the pointer `main` had immediately before the push and re-validates against both that and `docs/internal`'s current tip — on a simple staleness (the highest-frequency case) it fast-forwards and pushes; on anything else it opens a deduped `submodule-pointer-regression`-labeled issue instead. Guarded by a loop-prevention marker, a concurrency group, and a 7-day shadow-mode default (suppresses only the repair push, never the alert).
- `scripts/bump-docs-pointer.sh` — the canonical manual bump path every issue and error message points to. Ancestry-validates before leaving anything committed; reverts cleanly on failure.
- `scripts/tests/check-submodule-pointer-pr2609-replay.test.ts` — replays the real commits from the original incident through the ancestry engine (Wave 1), confirming it actually catches the true historical bug, not just a synthetic stand-in.

**Modified:**
- `.github/CODEOWNERS` — requires review on the one workflow file that can ever hold the write-capable credential, since this repo's branch protection has no PR-review requirement at all otherwise.
- `CLAUDE.md` — Monitoring & Alerts entry and a Troubleshooting row for the new GitHub Issue label.
- `docs/internal/adr/143-*.md` — records the invariant, why a CI job needs write access to `main` to close the gap fully, and states plainly what that credential will actually be capable of once provisioned (not just its intended use).

**Governance review fixes** (commit `2d9f56b15`, before this PR was opened):
- `pointer-autorepair` no longer requests `contents: write` on its default CI token — it was unused, since the real push uses a separate, narrower credential entirely.
- The repair-commit sequence now fails loudly instead of silently, closing a path where a mid-sequence error could have let the job proceed with stale content.

**Plan:** `docs/internal/implementation/smi-6260-docs-internal-pointer-regression-gate.md`
**Linear:** SMI-6260 (PR 2 of 2 — Wave 1, #2658, already merged)


wrsmith108:
PR Review complete — PR #2659 (SMI-6260 Wave 2: docs/internal ancestry gate + post-merge auto-repair)

Checks: 3 substantively executed (PR-07, PR-12, PR-14 all PASS), 11 SKIP by trigger absence, 0 findings, 0 fixes needed this round.
Verdict: PASS
Report: docs/internal/pr-reviews/2026-08-31-smi-6260-wave2-docs-internal-pointer-gate.md (committed to docs/internal on branch docs/smi-6260-wave2-pr-review, commit 6f091af, skillsmith-docs PR #946 — not yet merged into skillsmith's docs/internal gitlink, coordinator's call)

Same-family Claude fallback review (cross-model GPT-5.6-Sol dispatch blocked, SMI-6333) — a retry of a prior dispatch that died mid-run from an account rate limit, no partial work left behind.

Independently re-verified by direct code read (the single most safety-critical property in this PR): scripts/ci/submodule-pointer-autorepair.sh's case branch for every non-auto-repairable rule (R1/R5/R6/R7/R8/R11) calls only a GitHub-issue-opening helper, never git push. The sole push call is gated behind RULE=R3 AND non-shadow-mode AND a provisioned SKILLSMITH_MAIN_PUSH_PAT. Also re-verified both pre-PR governance-review fixes (2d9f56b15): unused contents:write grant removed, R3 repair-commit sequence now fails loudly instead of silently.

Known, pre-existing, non-blocking: this PR's own pointer-check job is FAILURE on itself (R8 — external contributors cannot bump docs/internal), because STRATEGY_SUBMODULE_PAT doesn't exist anywhere in this repo's CI (confirmed live, SMI-6337) — three existing workflows have been silently degrading on this same gap. pointer-check is simply the first consumer designed to hard-fail instead of degrading silently, working as designed. Not a required branch-protection check yet (Wave 2 Step 4 explicitly deferred pending SMI-6337). Every other CI check on this PR is green.

No fixes committed this round (nothing failed). PR is ready to merge from this review's standpoint — final merge decision is the coordinator's.

PR #2751, opened 2026-09-07
fix(hooks): SMI-6437 -- container self-heal verifies native-module health

wrsmith108:
## Business Summary

**What shipped:** the dev-container dependency self-heal (`skillsmith-dev-1`'s auto-`npm install` on push) now checks that native SQLite bindings actually work before letting a push through — either right after it declares success, or right after it fails. Previously, neither outcome checked this, so a broken binding from a failed self-heal could sit undetected until a completely unrelated later push happened to trip over it.

**Quality bar:** plan reviewed by GPT-5.6-Sol (cross-provider, via NEEDLE/ADR-128) before implementation; every finding independently re-verified against the live repo, not accepted on the reviewer's word alone. 35 test assertions across two test files, all passing against the real production scripts (fake Docker/npm, real lock/install logic).

**Found along the way:** the existing test file for this mechanism (`check-container-deps-fresh.test.sh`) had grown to 464 of the repo's 500-line limit; split into a shared fixture file plus a sibling test file for the lock-contention scenarios, following the existing `needle-dispatch-fixtures.sh` precedent — not filed separately, since it was a direct blocker for adding this PR's own test coverage.

**Net result:** fully shipped, no follow-up.

---

## Technical detail

**Root cause** (full writeup: `docs/internal/implementation/smi-6437-container-self-heal-native-verify.md`): a routine Dependabot bump (`@ruvector/core` 0.1.30→0.1.32) triggered the self-heal's `npm install`, which hit `EACCES` on a rename of a platform-mismatched optional dependency (`ruvector-core-linux-x64-gnu`, x64-only, on this arm64 container). The failed install left `packages/core/node_modules/better-sqlite3` broken (confirmed via a real functional test, not `require.resolve`). Ownership/ACL causes are ruled out with direct evidence; the exact EACCES trigger recurred once more, independently, on the same path during this investigation, and self-corrected on retry both times.

**Fix**: `scripts/lib/check-container-deps-fresh.sh` now invokes the existing native-module probe (`check-native-modules.sh`, SMI-5513, reused as-is) on both self-heal outcome paths:
- Success (`RC=0`, self-heal ran): a broken binding is now a hard failure instead of a silent "Self-healed" pass.
- Failure (`RC=4`, npm install itself failed): the existing error message now additionally names the restart remedy if bindings are also broken.

**Files**:
- `scripts/lib/check-container-deps-fresh.sh` — the two new call sites
- `scripts/tests/check-container-deps-fresh.test.sh` — trimmed to Scenarios 1-4 + 3 new native-probe scenarios
- `scripts/tests/check-container-deps-fresh-lock.test.sh` — new; Scenarios 5-9 (lock contention) moved here
- `scripts/tests/_lib/check-container-deps-fresh-fixtures.sh` — new; shared test setup
- `.github/workflows/validate-hooks.yml` — registers the two new test files
- `docs/internal/process/guards-and-opt-outs.md` — `SKILLSMITH_SKIP_NATIVE_CHECK` row updated to cover the new call sites

**CI note**: no new opt-out variable — `SKILLSMITH_SKIP_NATIVE_CHECK` (existing) silences the probe at both the original and new call sites.

### DIFF
--- a/.github/workflows/validate-hooks.yml
+++ b/.github/workflows/validate-hooks.yml
@@ -35,6 +35,8 @@ on:
       - 'scripts/lib/check-container-deps-fresh.sh'
       - 'scripts/lib/check-container-deps-fresh-inner.sh'
       - 'scripts/tests/check-container-deps-fresh.test.sh'
+      - 'scripts/tests/check-container-deps-fresh-lock.test.sh'
+      - 'scripts/tests/_lib/check-container-deps-fresh-fixtures.sh'
       - 'scripts/tests/session-mcp-disconnect-guard.test.sh'
       - 'scripts/tests/create-worktree-hooks.test.sh'
       - 'scripts/tests/worktree-hookspath.test.sh'
@@ -98,6 +100,8 @@ jobs:
             scripts/lib/check-container-deps-fresh-inner.sh \
             scripts/lib/check-hooks-path.sh \
             scripts/tests/check-container-deps-fresh.test.sh \
+            scripts/tests/check-container-deps-fresh-lock.test.sh \
+            scripts/tests/_lib/check-container-deps-fresh-fixtures.sh \
             scripts/tests/session-mcp-disconnect-guard.test.sh \
             scripts/tests/create-worktree-hooks.test.sh \
             scripts/tests/worktree-hookspath.test.sh \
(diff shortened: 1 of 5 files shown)


wrsmith108:
Reopening to re-trigger CI — no workflow runs fired on initial open (isolated missed webhook delivery, confirmed: other PRs opened around the same time triggered normally).

PR #2783, opened 2026-09-09
feat(mcp): SMI-6472 -- MCP server spec alignment (registry _meta, validation, annotations, structured output)

wrsmith108:
## Business Summary

**What shipped:**

- Skillsmith's listing on the official MCP registry will now actually carry its categories and keywords — they were being dropped on every publish because they sat in the wrong place in the manifest.
- `skill_validate` now catches two kinds of malformed skill it previously passed: a name that isn't a valid slug, and a name that doesn't match the folder the skill lives in. Both are requirements of the spec Skillsmith validates against.
- Every one of the 43 MCP tools now tells the client whether it only reads or can destroy something, plus a human-readable display name. Hosts can use this to auto-approve safe calls and confirm dangerous ones.
- The three most-used read-only tools (`search`, `get_skill`, `skill_validate`) now return machine-readable structured data alongside the existing text, so clients no longer have to re-parse free-form JSON and guess at its shape.

**Quality bar:** Three waves, each verified before commit — 3315 tests across 195 files, standards audit at zero failures, typecheck and lint clean, and a concurrency diff-audit per wave. Pre-merge review was dispatched to GPT-5.6-Sol (a different model family than the code's author, per ADR-128); it returned two should-fix findings, both fixed in this PR with regression tests, none deferred. Because this surface has no staging environment before it reaches npm and the public registry, it was also smoke-tested by driving the real built server with a real MCP client and calling the tools by hand.

**Found along the way:** four issues filed separately rather than folded in — a data-loss risk where one install path overwrites locally-modified skills without a backup (SMI-6477), and three repo-health problems that fail identically on `main` and are unrelated to this branch: `npm run preflight` red (SMI-6474), `npm run lint` red (SMI-6475). One tool's description claimed it was read-only while it wrote files; that one *was* fixed here, since shipping it next to a contradicting annotation in the same release would have been incoherent.

**Net result:** Ready for review. Not merged and not published — publishing to npm and the MCP registry is a separate, deliberately gated step for a human.

---

## Technical detail

Four commits on `feature/mcp-server-spec-alignment`, based on `5168db4ef`. Plan: `mcp-server-spec-alignment.md` under the internal implementation docs.

**Wave 1 — `27f8c0e9b`.** `server.json`'s `io.skillsmith/categories` + `io.skillsmith/keywords` renested under the reserved `_meta` key `io.modelcontextprotocol.registry/publisher-provided`; anything else under top-level `_meta` is dropped by the registry (confirmed live — `versions/latest` returned `_meta: {}`). New `audit:standards` **Check 66** guards regression, modelled on Check 53's field-length pattern, with 7 unit tests.

`skill_validate` gained a name-format check and `validateNameMatchesDirectory()`. Rather than a third copy of the name regex, the canonical `VALID_SKILL_NAME_RE`/`validateSkillName` moved to `@skillsmith/core`; `cli` re-exports, `mcp-server` imports from the root barrel it already imported from (zero new dependency edges). The vscode codegen sync was **retargeted, not retired** — retiring it would give the extension a `@skillsmith/core` dependency and trip ADR-113's explicit trip-wire.

`agent-pack.conformance.test.ts` now validates the pack in its **installed** layout (`<dir>/skillsmith-agent/SKILL.md`) rather than its source path. The asset lives under `src/assets/agent-pack/` by monorepo convention but always installs as `AGENT_PACK_SKILL_NAME`, so name and directory match everywhere it ships; validating the source path measured folder naming, not the shipped skill.

**Wave 2 — `1832de6fc`.** The load-bearing change is `index.ts`'s `ListToolsRequestSchema` handler, which mapped every tool to `{name, description, inputSchema}` and dropped everything else — annotating the schema constants alone would have been inert on the wire. It now passes `title`/`annotations` through, omitting each when absent, keeping the explicit allowlist so internal fields can't leak, typed against the SDK's real `ToolAnnotations`.

43 tools annotated: 21 read-only, 14 destructive, 8 additive-only. Multi-action tools take the worst case across every action. Classification was done by reading each handler through to its real filesystem/DB/network calls, not inferred from names.

Regression test drives a real server over stdio with a genuine SDK `Client` and asserts on the actual `tools/list` response — a unit test on the schema constants would pass even with the handler stripping fields. Negative control: with the passthrough disabled, 6 of 8 tests fail naming all 43 tools; `index.ts` restored byte-identical after.

**Wave 3 — `5d9c3582d`.** `outputSchema` on `search`/`get_skill`/`skill_validate`, `ok()` extended with opt-in `structuredContent`, same handler extended for `outputSchema`.

The SDK `Client` throws if a tool declares `outputSchema` but returns no `structuredContent` on a non-error path, and validates the payload against the schema. Each schema was derived from the response type in that tool's own module and cross-checked against its real return statements; there is exactly one non-error return path per tool, and every early/error return in the dispatch chain is `isError: true` (SDK-exempt). `ok()`'s new parameter is optional so the ~40 tools that don't opt in stay byte-identical; it remains stateless.

No `ajv` import — the plan's claim that it was already a dependency of `packages/mcp-server` is wrong (only `ajv-formats-draft2019` is declared). The SDK bundles its own validator, so the wire round-trip is the conformance test.

**Review fixes — `aea36f4a2`.** (1) `validateNameMatchesDirectory()` rejected valid skills given a relative path — `basename(dirname('SKILL.md'))` returns `'.'`, which never equals a valid name. Now resolves first; `resolve()` is lexical not realpath, so a symlink named after the skill still passes. (2) `structuredContent` diverged from the text block: the first-run-welcome and consent decorators rebuilt only `content[0].text`, so a structured-output client never saw `consent_required`/`privacy_url`. Both now route through one shared `spliceStructured()` helper, a no-op for tools without `structuredContent`.

**Two plan inaccuracies corrected during implementation**, noted for anyone reading the plan alongside the diff: the `ajv` claim above, and the plan's location of inline tool schemas in `audit-tool-dispatch.ts`/`provenance-tool-dispatch.ts` — those files contain no literals, only imports of six constants from `tools/*.ts`.

**Smoke pass** (real built server, real client): 43 tools returned with `title` + `annotations` on all; `uninstall_skill` reports `destructiveHint: true`; `outputSchema` advertised on exactly the three tools; `search` and `skill_validate` return both a text block and `structuredContent`, with the SDK validating against the declared schema and not throwing.

**Concurrency**: `concurrency-auditor` Mode B run per wave. Waves 1 and 3 clean. Wave 2's 6 hits and the review-fix commit's 1 hit are all pre-existing — function-local `Map`s in files touched only at the schema literal, and `consentCache`, a module-level Promise memo present on HEAD before this branch with its key derived at a single site.

**Deploy notes**: no migrations, no edge functions, no SQL. `packages/mcp-server/CHANGELOG.md` has one `[Unreleased]` entry per wave.


wrsmith108:
**Pre-merge review: PASS** (GPT-5.6-Sol via NEEDLE, ADR-128 — different model family than this PR's author)

| Check | Result |
|---|---|
| PR-01–06, 08, 09, 11, 13 | SKIP (no SQL, no new edge functions) |
| PR-07: Silent catch | PASS |
| PR-10: Export continuity | PASS — independently re-verified, no stale imports |
| PR-12: Plan vs diff | PASS |
| PR-14: audit:standards | PASS — 84 passed, 10 pre-existing warnings, 0 failed (independently re-run) |

0 findings. Full report: docs-internal PR [smith-horn/skillsmith-docs#1071](https://github.com/smith-horn/skillsmith-docs/pull/1071).

PR #2793, opened 2026-09-10
fix(security): SMI-6505 -- stop scoring an embedded key assigned a boolean as a credential

wrsmith108:
## Business Summary

**What shipped:** Skills that document FastAPI CORS setup can be installed again. The security scanner was treating `allow_credentials=True` — a configuration flag, not a credential — as a leaked secret, and a single such finding is enough to make the installer refuse the skill outright. Two skills in the live registry were affected.

**Quality bar:** Cross-model plan review (GPT-5.6-Sol via NEEDLE, ADR-128) rejected the first design; its central finding is what this implements. Two follow-up designs, including my own, were then discarded after testing showed each one let a real credential slip past the scanner. 1575 tests pass across 62 files. The scanner's two hand-maintained edge copies now have a behavioural parity test against the main implementation, which did not previously exist — the old check only compared the two copies to each other.

**Found along the way:** Four issues filed separately rather than folded in — **SMI-6508** (High: the mirror-image defect, where real secrets assigned to keys like `API_SECRETS` are *not* flagged today), **SMI-6506** (a pre-push check reports a misleading failure that masks the real one), **SMI-6507** (a hook prints a container command the docs forbid), and **SMI-6510** (credential-scanning prompts can't be sent for cross-model review, because the tool that carries them has its own credential scanner). Also corrected two factual errors in this issue's own suggested fix, and one in the plan doc, after measuring rather than reasoning.

**Net result:** Fully shipped for the reported case. Two formatting variants deliberately still flag, and one input class deliberately loses detection — all three are documented and pinned as tests, and the reasoning is in the plan. Live-population verification is the one open item.

---

### The fix

`CREDENTIALS_ASSIGN_PATTERN` (`/credentials\s*[:=]/i`) carries no left boundary, so it matches the `credentials=` inside `allow_credentials`. `isProseValue` then evaluates the single token `True`, which misses the stopword rule, the `>12`-token rule and the 2-token label carve-out alike, and falls through to the documented HIGH default.

So the discriminator is the **key**, and the key is available one frame up — `assignmentHasRealValue` already holds `line` and `heads[i].index`:

```ts
const keyIsEmbedded = idx > 0 && /[A-Za-z0-9_]/.test(line[idx - 1])
if (keyIsEmbedded && BOOLEAN_FLAG_VALUE.test(value.trim())) continue
```

with `BOOLEAN_FLAG_VALUE = /^(?:true|false)[\s,;)\]}]*$/i`.

### Why not a prefix boundary

`\b` does fix the reported case — this issue's own text says it would not, which is inverted; the *absence* of a word boundary is what makes `\b` refuse to match. The proposed negative lookbehind is identical to `\b` across all nine test cases.

Both are rejected because both stop matching `AWS_CREDENTIALS=hunter2`, a real credential. That false negative already ships for `secrets` (which does carry `\b`), so `API_SECRETS=hunter2` and `app_secrets=hunter2` are unflagged in production now — filed as **SMI-6508**. No uniform prefix change fixes both directions, so this PR changes no pattern.

### Why the regex is anchored at both ends

Two relaxations each fix a real formatting variant and each open a bypass. The value span runs to the next assignment key or EOL, so anything the regex accepts unexamined is where a credential can hide. Measured both ways:

| relaxation | fixes | opens |
|---|---|---|
| token-terminated | single-line multi-arg call | `my_credentials=true, Tr0ub4dor&3` → MEDIUM |
| trailing-comment branch | `allow_credentials=True,  # allow cookies` | `my_credentials=true # Tr0ub4dor&3` → MEDIUM |

Both bypass shapes are pinned as must-stay-HIGH tests. The two variants they would have fixed are accepted residuals.

### `SCANNER_RULESET_VERSION` → `2026-09-10.1`

Load-bearing. This is a severity-**lowering** change, so without the bump the `comparable` gate keeps reusing the stored pre-fix verdict and an already-scanned skill stays blocked. Same direction as SMI-5207's bump, opposite of SMI-6441's. The constant's own doc comment states the precedent.

`packages/mcp-server/tests/fixtures/security-audit/output-compat-baseline.json` is regenerated accordingly. Diff audited: only `rulesetVersion`, its three derived `key`/`acceptKey` hashes, and a two-candidate reorder (the test sorts by key, and the keys moved). Identical fingerprints and severities on both sides.

### Blast radius

Exactly 4 inputs change — 3 fixes plus `DB_PASSWORD=true` as the stated cost. Unchanged: bare `credentials:` / `secrets:` / `password:` keys, `AWS_CREDENTIALS=hunter2`, and every SMI-6441 fixture including `password: monkey dragon` (HIGH) and `password: horse staple` (MEDIUM).

### Files

| file | change |
|---|---|
| `packages/core/src/security/scanner/SecurityScanner.value-gate.ts` | the rule + `isEmbeddedKey`; module header amended (its "can never regress detection" claim no longer holds unconditionally) |
| `packages/core/src/security/scanner/patterns.ts` | ruleset version bump |
| `scripts/indexer/_shared/security-scanner-edge.value-gate.ts` | edge twin |
| `supabase/functions/_shared/security-scanner-edge.value-gate.ts` | edge twin (generated from the Node one, not hand-edited) |
| `packages/core/tests/security/sensitive-path-boolean-flag.test.ts` | new, 32 cases |
| `scripts/tests/indexer/security-scanner-edge.boolean-flag-parity.test.ts` | new, core↔edge behavioural parity |
| `packages/mcp-server/tests/fixtures/security-audit/output-compat-baseline.json` | regenerated |
| `packages/core/CHANGELOG.md` | `[Unreleased]` entry |

### Reviewer caveats

- **Do not "simplify" `BOOLEAN_FLAG_VALUE` to a token-terminated or comment-tolerant form.** Both look like improvements and both are measurably exploitable; the failing tests name the exact bypass.
- **The two edge twins must stay byte-identical.** `prettier --write` reformats the Node one but not the git-crypt-encrypted Deno one, so regenerate the Deno copy after any format pass rather than editing both. The `sed` one-liner is in the plan.
- **`security-audit.acceptance-candidates.test.ts` reads the built core**, so it gives a false pass on a source-only edit until `tsc --build` re-emits. It also lives under `packages/mcp-server/src/`, so a `packages/mcp-server/tests` glob skips it (SMI-6488).
- **`npm run preflight` is not the quality gate** — it exits 0 while printing three missing-dependency errors (the known SMI-6482 false positives) and runs no lint, typecheck, tests, or format.

### Open

Wave 2 live-population verification: confirm nothing moves MEDIUM→HIGH, confirm the two blocked rows install, and count the `DB_PASSWORD=true` and kebab-case (`allow-credentials=true`) classes. If the accepted cost is common in the population the trade changes, and that should be settled before merge.

The cross-model **confirmation** round on the final design has not run — Codex hit its usage limit. The design is verified by measurement, not by an independent second pass.

Plan: `docs/internal/implementation/smi-6505-credentials-assign-false-positive.md`

Closes SMI-6505.


wrsmith108:
## Wave 2 — live-population verification (the pre-merge gate's remaining blocker)

Run against prod, 489,521 skills. This closes the cross-family review's one blocking finding: the commit's "exactly 4 inputs change" described the fixture matrix, not the population.

### The population this change can move

| class | count |
|---|---|
| gate trigger (any gated keyword assignment) | 8 |
| **embedded keyword + boolean — the entire set this change can move** | **2** |
| `allow_credentials` specifically | 2 |
| `DB_PASSWORD=true` — the accepted-cost class | **0** |
| kebab-case (`allow-credentials=true`) residual | **0** |

Both movable rows are `abenezer101/security-fastapi` (duplicate entries, `security_score = 0`) — exactly the skill named in SMI-6505's evidence.

**The accepted cost is zero rows in practice.** The plan said a nonzero accepted-cost population would change the FP/FN trade and should be settled before merge; measured, it is empty. The kebab-case residual is also empty, so that scoping decision costs nothing live.

### Nothing moves MEDIUM to HIGH

Two independent grounds:

1. **Structural.** The new rule only adds a `continue`, so `assignmentHasRealValue` can only return `true` *less* often. Severity is monotone non-increasing under this change, so a MEDIUM to HIGH transition is unreachable by construction — not merely unobserved.
2. **Empirical.** The movable set is 2 rows, and both are the intended fix.

### The two rows now install — verified against the real stored bytes

The exact live line, extracted from the rows rather than reconstructed:

```
  - user: "Check FastAPI CORS setup" → verify origins when allow_credentials=True
```

Run through the shipped gate and the real scanner:

```
assignmentHasRealValue        -> false   (MEDIUM)
high/critical sensitive_path  -> 0
report.passed                 -> true
```

`report.passed` is the exact value `skill-installation.service.ts` rejects on, so this is the install gate itself clearing, not a proxy. Confirmed identical with and without leading indentation.

### Method note

The SMI-6441 blast-radius harness was **not** reused. It is built around that wave's `weakPasswordVeto: false` counterfactual seam, and this change has no equivalent toggle — so its before/after model does not apply. Rather than add a production option purely to enable a harness, the "before" here is established structurally (the monotonicity argument above) and the population is measured directly.

PR #2797, opened 2026-09-11
docs: SMI-6493 follow-up implementation plans (6511, 6512, 6513, 6517)

wrsmith108:
## Business Summary

**What shipped:** Four implementation plans for the SMI-6493 follow-up work — SMI-6511 (registry assertion before publish), SMI-6512 (unguarded `mktemp` red-failing a verified publish), SMI-6513 (byte-identity check across publish.yml's four verify blocks), SMI-6517 (pre-commit guard for backward submodule pointers). Plus the regenerated root folder index.

**Quality bar:** Two of the four plans went through adversarial plan-review. Both returned APPROVE-WITH-CHANGES with blocking findings, and SMI-6511's plan was revised against all twelve. Every load-bearing claim was re-verified against the live repo rather than accepted from the drafting agent — that caught a wrong conclusion in each of the two reviewed plans. No code ships here; all four remain ADR-109-gated behind plan-review.

**Found along the way:**
- SMI-6516 — reproduced on demand (five clean runs, degradation on the sixth) after being filed on an unverified correlation. Handed to the session consolidating the container surface, along with the repro.
- SMI-6520 — a repair script reporting a green count for unverified symlinks. Same session now owns it; the key constraint (the escaping symlink is load-bearing, not a defect) is recorded so nobody "fixes" it.
- SMI-6513 — `audit:standards` has two exit paths that disagree: `fail()` reds a full run but exits 0 under `--only` unless the check also returns false. Recorded as a hard requirement on the issue.
- Two template-conformance nits on the 6512 and 6513 plans, filed onto their pending revision passes rather than churning the submodule pointer for heading placement.

**Net result:** Plans are committed and reviewable. **None is ready to implement.**

> ### ⚠️ Merging this PR registers documentation. It does NOT approve implementation.
>
> Every one of the four plans is explicitly not-yet-implementable. Do not start work from any of them without clearing the gate named below:
>
> | Plan | What blocks implementation |
> |---|---|
> | **SMI-6511** | Two BLOCKERs found in review. Revised, but the **SMI-6015 precondition is unmet**: BLOCKER 2's collector reorder needs its own narrowly-scoped adversarial round before any code is written. |
> | **SMI-6517** | **Not plan-reviewed at the time of this commit.** Review has since found a BLOCKER — the `old_mode != 160000` parse rule never matches, because `git diff --cached --raw` emits a leading colon, so the guard would be inert. The committed doc is the pre-revision version. |
> | **SMI-6513** | No plan-review yet. Six open questions, including whether it needs its own ADR. |
> | **SMI-6512** | No plan-review yet. **Must not be implemented before SMI-6513** — it bumps a digest SMI-6513 defines. That ordering is prose-only, with nothing enforcing it. |

---

## Technical detail

**Changed path:** `docs/internal` gitlink only — `eafd4ca` → `2860807`.

**Submodule commits registered:**
- `13fd193` — the four plan docs (1,986 lines)
- `2860807` — regenerated root `index.md` Folder Reference count, 695 → 699

**Why the index regen is in scope:** `docs-folder-index.yml` triggers on the literal `docs/internal` gitlink path (SMI-5523 added it precisely because `docs/internal/**` never matches a pointer bump), so a pointer-only commit fires that required check. `gen-docs-folder-index.mjs --check` returned stale by exactly four before the fix and passes now. Re-verified immediately before push, since the count races between concurrent sessions.

**Bundling rationale:** one submodule commit and one pointer bump rather than four. Four separate bumps serialize badly — each merge invalidates the next one's pointer, which is the SMI-6502 mutual-starvation shape.

**Ancestry:** forward and non-diverged. `eafd4ca` is an ancestor of `2860807`, the reverse is false, and `2860807` is on the submodule's `origin/main` — so no orphaned gitlink (ADR-143 / SMI-6260 R1).

**Governance:** `audit:standards` exit 0 — 83 passed, 11 warnings, 0 failed, with zero new failures attributable to these commits. No secrets, no personal names. The 500-line file-length limit does not apply to markdown (verified at both the pre-commit gate and CI Check 3), so the 816- and 602-line plans need no ignore-list entry.

**CI caveat:** this PR is docs-only with SMI refs throughout, which is exactly the shape `verify-implementation.ts` fails — hence `[skip-impl-check]` below.

**Pre-merge gate:** run, verdict MERGE WITH CONDITIONS. 12 checks SKIP (no SQL, migrations, RLS, edge functions, TS, exports, or enums in a gitlink-only diff), PR-12 and PR-14 PASS. PR-14 was independently re-run (83 passed / 11 warnings / 0 failed) and confirmed structurally insensitive to these files — no check in `audit-standards.mjs` enumerates `docs/internal/implementation/*.md`.

**Review provenance — ADR-128 deviation, disclosed not buried.** This gate was **NOT** run cross-family. It is meant to dispatch to GPT-5.6-Sol via `scripts/needle/dispatch.sh`; NEEDLE failed twice tonight (`outcome=success-without-agent-message`). The gate ran same-family Claude as an explicitly authorized fallback, which carries a correlated-blindspot weakness — it may share the authoring session's framing errors. Worth re-gating through NEEDLE once it recovers. Datapoints are on SMI-6510.

Correction worth recording: my earlier claim that `dispatch.sh` "exits 0 while failing" was wrong. It correctly exits 1 on `success-without-agent-message`; my invocation piped it through `tail`, which reported `tail`'s status instead. The tool reported the failure accurately.

[skip-impl-check]

https://claude.ai/code/session_01LjPm6wr3yqA4KrhRJnNv2E


wrsmith108:
## Closing as superseded — the work is already on `main`

Not abandoned. **All four plans landed on `main` via #2800**, which merged first (squash `7edec3833`). This PR is now a genuine no-op: its gitlink and `main`'s are byte-identical, so merging would produce an empty commit.

Verified by ancestry rather than assumed:

```
main registers docs/internal:  76e540593
this PR's head registers:      76e540593   <-- identical

13fd193 (four plan docs)      -> ancestor of main's pointer: YES
2860807 (folder index regen)  -> ancestor of main's pointer: YES

implementation/smi-6511-publish-registry-assertion.md      present
implementation/smi-6512-smokedir-guard.md                  present
implementation/smi-6513-verify-block-byte-identity.md      present
implementation/smi-6517-precommit-gitlink-backward-guard.md present

Folder Reference count at main's pointer: 707
```

Both of my submodule commits are ancestors of what `main` now registers, all four plan docs are present at that pointer, and the folder-index regeneration is included. Nothing in this branch is lost by closing it.

### Why it worked out this way

The gitlink is repo-wide, so whichever PR landed first was always going to carry the other's work. The other session verified up front that `13fd193` was already an ancestor of their tip — meaning we were never competing for content, only for the pointer. Once that was established, the merge order stopped mattering.

### What this cost, recorded for SMI-6502

This PR went through five `pointer-check` R3 / `BEHIND` cycles before the contention was resolved by the two sessions agreeing to serialize, not by either winning a race:

```
re-bump -> 2860807   R3: behind by 1   (f135e10 landed inside the CI window)
re-bump -> f135e10   R3: behind by 1   (76e5405 landed inside the CI window)
```

Measured on `docs/internal` main over 199 commits: p50 inter-commit gap **12.4 min** (p25 4.8) against roughly **11 min** of full CI. At that ratio each attempt is close to a coin flip, so this is the expected outcome rather than bad luck — which is what makes the honest framing "R3 is unsatisfiable under concurrency" rather than "R3 is redundant."

`--admin` was never used. It would have bypassed a genuinely failing required check and left `main`'s pointer R3-stale; with `SKILLSMITH_MAIN_PUSH_PAT` unprovisioned, `pointer-autorepair` cannot fast-forward and degrades to opening a `submodule-pointer-regression` issue — pushing the cost onto a shared alerting surface.

### Status of the plans themselves — unchanged by this closure

**None is ready to implement.** SMI-6511 has an unmet SMI-6015 precondition; SMI-6517 has a BLOCKER (its parse rule never matches, so the guard would be inert); SMI-6513 and SMI-6512 have not been plan-reviewed. Merging documentation was never approval to implement, and closing this PR changes nothing about that.

PR #2813, opened 2026-09-12
docs(plans): SMI-6511 + SMI-6512 + SMI-6513 + SMI-6517 -- revise four plans against cross-family review

wrsmith108:
## Business Summary

**What shipped:** Four implementation plans for the `publish.yml` follow-up work are now reviewed and corrected. They were written but never reviewed, and the review found real defects in three of them.

- SMI-6517's proposed guard would never have fired. It compares a field that git emits with a leading colon, so every case it exists to catch was classified as a different case and skipped.
- SMI-6512 and SMI-6513 would have broken each other. SMI-6512's fix rewrites the exact line SMI-6513 uses to find its place in the file, and SMI-6513 stops rather than guess — so landing both as written would have left a permanently failing build check with no way to clear it.
- SMI-6511 was checked a second time specifically because its earlier fix was the kind that tends to introduce a second, subtler version of the same bug. It came back clean.

**Quality bar:** Every review ran on a different model family than wrote the plans (GPT-5.6-Sol, per ADR-128) rather than the same-family default. Claims were executed rather than reasoned about: git's own output format, shell behaviour under the real flags, the anchor regexes, and the build-check numbering were each run against a case table before being written down. Two findings from the reviews were themselves wrong and are recorded as disproved.

**Found along the way:** Four things, filed separately rather than folded in — SMI-6568 (a commit-time safety check demands a container it doesn't need, because its own exemption can't apply to a branch's first commit), SMI-6569 (an unexplained check result, filed as evidence with the cause explicitly not claimed), plus corrections to two facts that had gone stale since they were last verified: the next free build-check number, and which return value from a check actually reports failure.

**Net result:** All four plans are ready to implement, in a fixed order: SMI-6513 first, then SMI-6512. No code changed — these are plans, gated by ADR-109. SMI-6493 itself stays open; only a real publish run closes it, and none has run since its fix merged.

---

## Technical detail

One commit, one file changed in the outer repo: the `docs/internal` gitlink, `8f833bf` → `d154349`.

`d154349` carries the four revised plans plus the sibling session's `fedefc4` (SMI-6564). Forward-only, verified by hand before committing: `8f833bf` is a strict ancestor of `d154349`, and the reverse ancestry test fails as it should. No sibling commit is dropped.

### The anchor conflict (SMI-6512 / SMI-6513)

SMI-6513 splits `publish.yml`'s block 4 at `/^SMOKE_DIR=/` and, per its own spec, fails when that matches anything other than exactly once. SMI-6512's fix rewrites that line to `if ! SMOKE_DIR="$(mktemp -d)"; then`. Measured against the post-fix string, all three candidate anchors fail:

```
/^SMOKE_DIR=/m                      -> false
/^\s*SMOKE_DIR=/m                   -> false
/^\s*SMOKE_DIR="\$\(mktemp -d\)"/   -> false
```

SMI-6512's stated remedy — bump the pinned digest — does not apply, because the split never succeeds and there is no digest to bump.

Resolved by an explicit `# audit:verify-block-smoke-tail` marker, introduced by SMI-6513, located in the parsed `run` string **before** normalization strips whole-line comments, and locked by a test that asserts the marker is missing when given a comment-stripped body. A loosened regex was rejected: it encodes two transient implementations rather than the boundary, so `mktemp --directory`, `readonly SMOKE_DIR=`, respacing, or a helper extraction each break it again.

### SMI-6517

`git diff --cached --raw` emits field 1 with a leading colon:

```
modify: f1=[:160000] f2=[160000] f3=[f7f7b5540] f4=[8f833bfdc] f5=[M]
add:    f1=[:000000] f2=[160000] f3=[0000000]   f4=[cb84b35]   f5=[A]
```

The spec compared `old_mode != 160000`, which is true for `:160000`, so every gitlink modification was classified as an ADD and skipped. Fixed by stripping the colon and testing positively. The zero-SHA alternative was rejected on two measurements: `--raw` abbreviation width is variable (7 chars in a fresh repo, 9 here — `core.abbrev` scales with object count), so any fixed-width zero comparison is wrong on a differently-sized repo; and it misses type-change rows (`:100644 160000 <blob> <commit> T`).

Two of the six MAJORs invert on measurement:

- A clean `git revert` does not run `pre-commit` at all — reproduced independently in two fixtures, hook fired zero times across a revert that moved a gitlink backward. The guard cannot block the documented rollback path, so this is a coverage hole, not a false positive, and no escape hatch is needed.
- `--amend` cannot be detected: args, all `GIT_*` env, and the git-dir listing are byte-identical between a normal commit and an amend (`diff` exit 0), and the one candidate heuristic was falsified by an ordinary commit made in the same second as its parent.

Scope narrowed to block `docs/internal` and warn on the three strategy mounts, resting on the R5-rewind false-positive class alone. A second supporting argument was withdrawn after measurement showed it fails open and applies to all four mounts equally.

### SMI-6511

The SMI-6015 confirmation round over the collector reorder returned APPROVE — no second ordering defect, the second pass needs no loop (a scope selects a key whose value is a URL, never another scope), the queried key set is complete, and no fail-open path is introduced. One MINOR applied: case 34 exits 0 `via: package-scope` under both the old and new ordering, so it guards the reorder rather than catching the original defect.

### Check numbering

Checks 69 and 70 were taken by SMI-6515 and SMI-6518 in `014b3acda`, after "next free is 69" was verified. Renumbered to 71, and all three affected plans now carry the command to re-derive the number rather than the answer, since it will rot again before implementation. SMI-6494 tracks the collision hazard.

### Review disclosure

Reviews were dispatched to GPT-5.6-Sol via `scripts/needle/dispatch.sh`, satisfying ADR-128. NEEDLE had been believed unusable; the actual failure was the SMI-5709 pre-flight guard rejecting an ordinary 52-character plan-doc path under its 44-character heuristic. Applying SMI-6510's documented workaround — name the directory prefix once in prose, bare filenames after — made all three dispatches succeed.

### Bypass disclosure

The pointer-bump commit used `SKILLSMITH_WORKTREE_PREPUSH_HARDFAIL_DISABLE=1`, scoped to that one commit, with explicit authorization. The staged change is a single `docs/internal` gitlink bump, which the repo's own `is_docs_only()` classifies as docs-only; the pre-commit gate reads the committed list instead, which is empty on a branch's first commit, so its own carve-out could not fire. Filed as SMI-6568. The push needed no bypass — the carve-out works correctly at pre-push, confirming the defect is scoped to pre-commit.

### Pre-merge gate (cross-family, ADR-128)

Run via GPT-5.6-Sol through NEEDLE, not the skill's default same-family Claude path. Checks 1–11 and 13 SKIP — no SQL, TypeScript, enum, edge-function surface or module move in the diff. Plan-vs-diff PASS (the plans *are* the diff; none falsely claims to have shipped). Standards-audit delta PASS (no executable code changed). State-flip SKIP — the plans describe future flips; the diff performs none.

Four findings. Three fixed in `001f83c`:

- **MAJOR** — SMI-6512's checklist promised "no bare, non-annotated non-zero exit remains" after the fix. False the moment it was ticked: `sleep 8` in the retry loop is still bare, and under `bash -e` a failing command aborts silently (`bash -e -c 'false; echo REACHED'` → no output, exit 1). Narrowed to explicit exit paths, with the omission stated rather than implied.
- **MINOR** — the two coupled plans defined different tail spans. SMI-6513 excludes the marker line from both regions and matches by trimmed whole-line equality; SMI-6512 said "from that marker line through the end," which includes it and permits substring matching. Aligned exactly, with the near-miss non-matches named.
- **MINOR** — SMI-6517's conditional index obligation answered rather than hedged (`implementation/` has no `index.md`). Found by the `/governance` pass.

**One finding accepted, its remedy rejected.** The reviewer correctly observed that the pointer bump also carries `fedefc4` (SMI-6564 — a parity opt-out plan and a guards-registry row), which this PR does not otherwise describe. Disclosed here. But it proposed re-basing the pointer onto a submodule commit containing only this PR's content, and **that must not be done**: it would move the gitlink backward past a sibling session's commit, which is the R7 regression ADR-143 exists to prevent. Forward-only is the invariant; carrying a sibling's commit is the correct behaviour, not scope creep.

[skip-impl-check]


wrsmith108:
Closing as **superseded** — main already registers everything this PR would.

This PR bumps `docs/internal` to `001f83c62`. Main now registers `be0852e07`, and `001f83c62` is a strict ancestor of it — verified with `git merge-base --is-ancestor`. A sibling session's pointer bump carried this content onto main ahead of me.

Confirmed present on main at `be0852e`: the four revised plans (SMI-6511 / 6512 / 6513 / 6517), the three tier-3 plans (SMI-6497 / 6498 / 6499), and ADR-156. Nothing in this branch is lost by closing it.

Worth recording why this happened, since it is a repeatable mistake rather than bad luck: I split docs work across two PRs, each pinning a **specific** `docs/internal` commit. A pinned pointer is fragile by construction — any sibling merge that advances main past it turns the PR into a backward (R7) or diverged (R6) regression through no action of mine. Both of these PRs were invalidated that way, twice, by two different sessions. One pointer bump to the live tip would not have been.

The remaining work from this cluster is the **tier-3 plan revisions**, which address three BLOCKERs a cross-family review found in the versions currently on main. Those land as their own change against the current tip rather than through either of these branches.

PR #2818, opened 2026-09-12
docs(plans): SMI-6497 + SMI-6498 + SMI-6499 -- tier-3 publish.yml follow-up plans + ADR-156

wrsmith108:
## Business Summary

**What shipped:** Plans for the last three `publish.yml` follow-ups, plus one architecture decision record. No code changed — these are gated plan-only deliverables.

- **SMI-6497** — the four copy-pasted post-publish verify loops get extracted into one shared shell helper. The plan picks a `scripts/ci/` helper over the conventional composite action for a measured reason, recorded as ADR-156.
- **SMI-6498** — the enterprise package has no post-publish check at all, and a separate smoke job waits a fixed 15 seconds before doing a real install. SMI-6493's fix made that second one *worse*, not merely left it latent.
- **SMI-6499** — a failure anywhere in the publish run suppresses the public mirror sync, even for a package that published fine. Two real incidents, both over an hour to manual repair.

**Quality bar:** Every claim that could be executed was executed before it entered a plan — job structure, regex predicates, shell exit semantics, registry probes, and both candidate designs driven through the same case table. Three separate findings were *rejected* on measurement rather than written up, including one of my own.

**Found along the way:** SMI-6576 — the root `CHANGELOG.md`'s `[Unreleased]` section is never drained, so release-cadence's "is there anything to release?" gate is permanently open. Filed separately.

**Net result:** All three ready for review. **This PR must merge after #2813** — see below.

---

## ⚠️ Merge order

**PR #2813 must merge before this one.**

This PR registers `docs/internal` at `b219d29`, which is strictly ahead of #2813's `001f83c`. Merging this first would leave #2813 setting a pointer that is *backward* relative to `main` — the R7 regression ADR-143 exists to prevent.

```
8f833bf  (main today)
  └─ 001f83c  <- PR #2813
       └─ 198c161  (sibling session, SMI-6516)
            └─ b219d29  <- this PR
```

Verified by hand: `001f83c` is a strict ancestor of `b219d29`, and `8f833bf` is a strict ancestor of both. The sibling's `198c161` is carried, not dropped.

---

### SMI-6497 — why a shell helper, not a composite action (ADR-156)

A composite action is the conventional choice here; `unlock-git-crypt` has 24 call sites across 7 workflows. It was rejected on a measurement:

```
$ grep -rln "github/actions" scripts/tests/
scripts/tests/audit-workflow-sha-pin.test.ts        <- SHA-pin static audit
scripts/tests/verify-implementation.test.ts         <- path classification
scripts/tests/fixtures/workflow-sha-pin/…yml        <- fixture
```

No composite action in this repo has ever had its **shell executed** by a test, because a composite action's shell has no local entry point. Against that, 25 `scripts/tests/*.test.sh` files exist for shell helpers. This matters specifically because SMI-6513's own diagnosis is that the original defect survived a green suite for want of a test seam — a remedy shipping without one repeats the error.

**Unify rather than parameterise.** The `break`-plus-post-loop-guard shape in the `skillsmith-cli` block exists for exactly one reason: the smoke loop must run in the same `run:` block. Promote the smoke tail to its own step and the difference disappears. Both the current shape and the proposed two-step shape were driven through the same six cases — identical on all six rows, including the two defect rows. A `--no-exit` flag would have preserved a distinction with nothing left to distinguish.

**SMI-6513's check is deleted, not narrowed**, since full consolidation leaves one block. Its replacement protects what becomes non-structural — that the four *call sites* invoke the helper identically — as a vitest case rather than another `audit:standards` check.

### SMI-6498 — the enterprise gap and the `sleep 15`

Verify-step census, by parsing the workflow:

```
publish-core            Verify core on npm
publish-mcp-server      Verify mcp-server on npm
publish-cli             Verify cli on npm
publish-skillsmith-cli  Verify skillsmith-cli on npm + smoke the wrapper
publish-enterprise      NONE
```

The enterprise step is deliberately **not** a copy of the other four: GitHub Packages requires a token even to read, and npm's ~5-minute publish-time malware scan — SMI-6493's entire budget rationale — is npmjs-specific. So it gets a smaller, separately justified budget. Its name matches SMI-6513's discovery predicate on purpose rather than dodging it, which means registering it is a required step, not an afterthought.

On the smoke job: `scripts/smoke-test-published.ts:94` runs a real `npm install`, which needs the tarball, not the packument. Before SMI-6493, a slow publish red-failed at verify and `smoke-test` never ran. After, it passes verify and hits a fixed 15s wait with no retry budget at all. The blind sleep is deleted in favour of retrying the install itself.

### SMI-6499 — and a pattern deliberately not written

Two real incidents, verified from the run records:

```
2026-09-05 and 2026-09-09, publish.yml:
  success  Publish @skillsmith/core
  success  Publish @skillsmith/mcp-server
  success  Publish @smith-horn/enterprise
  failure  Publish @skillsmith/cli     <- only failure, both runs
  skipped  Publish skillsmith-cli
```

Four packages published; one job failed; the whole run reported `failure`; the mirror sync was suppressed for mcp-server, which had published fine. Recovery took 1h13m and 1h27m of manual dispatch.

**No `concurrency-patterns.md` entry was written.** SMI-6499 was evaluated against SMI-6502's shape and ruled out: SMI-6502 requires an uncoordinated concurrent actor and elapsed time, whereas all 11 publish jobs are finished when the single `workflow_run` event fires — a static aggregation-granularity mistake, reproducible from one completed run. Recorded as "evaluated and ruled out," matching the existing Pattern 5 precedent, rather than forcing an entry the auditor would then match on.

### Hard landing order for implementation

**SMI-6513 → SMI-6512 → SMI-6497.** Landing 6497 first makes SMI-6512's marker precondition (`grep -c` for the split marker) return 0, which its own plan reads as "6513 hasn't landed" — a wrong diagnosis pointing at the wrong remedy.

### Not verified, stated rather than glossed

The mandated P-7 state-flip scanner is absent at this worktree's strategy-submodule pin; a hand-rolled substitute was used and labelled as such, with the real scanner required at implementation time. The "GitHub Packages can take up to an hour" community claim is cited in-repo but was never independently confirmed. Neither SMI-6513 nor SMI-6497 exists as shipped code, so both plans treat the other's shape as a value to re-check.

[skip-impl-check]


wrsmith108:
Closing as **superseded** — main already registers everything this PR would.

This PR bumps `docs/internal` to `1fbc1e288`. Main now registers `be0852e07`, and `1fbc1e288` is a strict ancestor of it — verified with `git merge-base --is-ancestor`. A sibling session's pointer bump carried this content onto main ahead of me.

Confirmed present on main at `be0852e`: the four revised plans (SMI-6511 / 6512 / 6513 / 6517), the three tier-3 plans (SMI-6497 / 6498 / 6499), and ADR-156. Nothing in this branch is lost by closing it.

Worth recording why this happened, since it is a repeatable mistake rather than bad luck: I split docs work across two PRs, each pinning a **specific** `docs/internal` commit. A pinned pointer is fragile by construction — any sibling merge that advances main past it turns the PR into a backward (R7) or diverged (R6) regression through no action of mine. Both of these PRs were invalidated that way, twice, by two different sessions. One pointer bump to the live tip would not have been.

The remaining work from this cluster is the **tier-3 plan revisions**, which address three BLOCKERs a cross-family review found in the versions currently on main. Those land as their own change against the current tip rather than through either of these branches.

PR #2819, opened 2026-09-13
docs(adr): ADR-153 -- worktree container mount topology, cause left unestablished

wrsmith108:
## Business Summary

**What shipped:** ADR-153 writes down how a worktree's dev container actually reaches its dependencies — through a symlink that resolves *outside* the container's own working directory — and records that mounts nested beneath it silently detach while the container runs. Both facts have misled people twice before; neither was written anywhere a reader would find.

**Quality bar:** Every measurement is attributed and kept separate from the hypotheses. The cause of the detachment is stated as **unestablished**, in those words, rather than filled in with the most plausible story.

**Found along the way:** The leading hypothesis this wave started with now has evidence against it. Recorded as such, alongside two alternatives, none asserted.

**Net result:** Done. This is the last deliverable the SMI-6516/6520 plan allocated but deliberately left uncreated.

---

## Technical detail

Three superproject commits, no code, advancing the `docs/internal` gitlink from `be0852e07` (what `main` registers) to `f87c529` — the submodule's own `main` tip. `49e6aca78` for the ADR itself, `d519c8174` for a count correction inside it, `33895596e` to land the pointer on the tip. Adds the ADR, its `adr/index.md` row, and the regenerated Folder Reference count (adr 85 → 86).

**The bump necessarily carries five commits that are not this PR's work** — other sessions' plan revisions, a pr-reviews index row, a retro, and ADR-156, all already merged on `docs/internal`'s own `main`. A gitlink bump advances along the submodule's main; it cannot cherry-pick. Landing on the tip is not optional either: the required `pointer-check` status enforces R3, which failed this PR at `e69d87d66` with *"stale: behind `smith-horn/skillsmith-docs:main` by 4 commits"*. Re-bumped with `scripts/bump-docs-pointer.sh`, which ancestry-validates before it commits.

Ancestry verified in both directions, and the tip verified against the remote rather than a local ref:

```
$ git ls-remote https://github.com/smith-horn/skillsmith-docs.git refs/heads/main
f87c529dc3246d8964444bee0a7375946a4d0a8a	refs/heads/main
```

Worth recording because the local warn-only pre-push check disagreed: it reported `R1: f87c529... was never pushed, or its branch was deleted`. That is a false negative from a stale local view — the object-existence-vs-divergence trap, where an unfetched object reads exactly like a diverged one. `ls-remote` is the arbiter, and CI's own `pointer-check` is the gate; the local warning is neither.

### What it records

A worktree container bind-mounts the worktree at `/app`. Each per-package `node_modules` symlink is sized `../../../../` from the **host's** nesting depth — correct on the host, and the point of SMI-4381. Inside the container `/app` *is* the worktree, so four levels up clamps at `/` and the link resolves to `/packages/<pkg>/node_modules`, outside `/app` entirely. Docker's `mount(2)` follows symlinks resolving a bind destination, so **~165 declared destinations** land in the escaped namespace.

The container works because every declaration and every consumer agree on that namespace — not because the escape was designed.

### Three decisions

1. **The depth literal is not to be changed in isolation** — ~165 destinations resolve through the clamp. Stated narrowly on purpose; it is not a claim the current declarations are correct in any deeper sense.
2. **A coordinated redesign is explicitly not excluded.** The plan's review constraint required documenting alternatives without canonising `..` clamping as the only architecture.
3. **Verification asserts composition, never a total** — a normalized declared-vs-mounted destination-set diff. This is ADR-151 applied to mounts.

Decision 3 is a decision rather than a detail because of one measurement: a degraded container showed **164** mount lines against a healthy **165**, while **32 declared destinations were missing**. The counts stay close because ~130 surviving mounts sit under `/node_modules` and dominate the total. The plan's own first-draft check asked for a "byte-identical 165-line comparison" and was removed on that evidence.

### The finding that points away from the obvious culprit

**The main checkout runs a structurally different topology and shares the fault.** No symlink, no clamp, whole-directory per-package volumes — and the identical Mach-O fall-through, byte-size identical at 1,914,736 B, with 8 of 9 declared volumes detached.

So the invariant is about **nested-mount durability**, not about the symlink. The same detachment also unifies two symptoms previously tracked apart: the broken native binding and SMI-6520's `EROFS` on `.vite-temp` are one event.

### Cause: unestablished, and said so

Measurements are tabulated with attribution. Hypotheses are listed separately, none asserted:

- `create-worktree.sh` Step 7's host-side symlink sync — **the leading hypothesis when SMI-6516 was written, now with evidence against it.** A peer bracketed a single ~10-minute verification window in which a container went ELF → Mach-O with no restart, no recreate, no npm and no repair script; a separate run confirmed `repair-worktrees.sh` halts at its active-container guard before the native rebuild.
- Something reaching the host native-rebuild path without that guard.
- Directory replacement under a live mountpoint (npm reify, SMI-6378 shape).

A clean experiment remains available and nobody has run it. That is recorded as an open gap rather than closed with the most attractive of the three.

### Scope held

Does **not** absorb SMI-6524 (git unusable in worktree containers) or the SMI-6488 enforcement design. Those get their own ADRs.

### Verification

- `node scripts/gen-docs-folder-index.mjs --check` → `✓ Folder Reference table is up to date`
- The final pointer was **not** hand-staged. The first two bumps were, after confirming the submodule was on `main` at its tip — hand-staging from a worktree on a *feature branch* is what produced #2809's off-main pointer earlier today, so this one was checked first. The third used `scripts/bump-docs-pointer.sh`, which resolved `e69d87d66 → f87c529` and ancestry-validated before committing. The SMI-6527 caveat — that the script reports "nothing to do" while HEAD still registers something stale, because it compares the working tree to its target rather than what HEAD registers — did not bite here: the working tree genuinely differed from the target, so the script had real work to do. It remains a live trap for the case where it doesn't.
- 7 submodule commits are spanned in total, 2 of them this PR's (`99e43c0`, `e69d87d`) — verified with `git rev-list --count be0852e07..f87c529`, not counted by eye.
- Markdown line-length (`MD013`) is not enforced on these files — ADR-151 and ADR-155 carry 32 and 87 of the same warnings respectively, and CI's Markdown Lint job runs `.markdownlint.json` against superproject files, not the submodule.

[skip-impl-check] — docs-only. One superproject commit bumping the `docs/internal` gitlink; the content is an architecture decision record inside the submodule. There is no source code to add.


wrsmith108:
Closing: **ADR-153 is already on `main`.** Not abandoned — landed by another route while this PR was waiting on its pointer race.

`main` now registers `docs/internal` at `c9bb93e`, bumped there by #2825 (SMI-6513). That pointer contains this PR's entire deliverable:

```
$ git ls-tree origin/main docs/internal
160000 commit c9bb93e5309e13b7ffd8e6988e6487ef1d26e08a	docs/internal

$ git -C docs/internal rev-parse c9bb93e:adr/153-worktree-container-node-modules-mount-topology.md
8b4508e1344ce93340824cecd3f6fd71f998c1b2      # byte-identical to this PR's blob

$ git -C docs/internal merge-base --is-ancestor 99e43c0 c9bb93e   # ADR-153 itself      -> contained
$ git -C docs/internal merge-base --is-ancestor e69d87d c9bb93e   # its count correction -> contained
```

The `adr/index.md` row and the Folder Reference count (adr 85 → 86) came along with it. So merging this PR would add nothing: re-bumped to `c9bb93e` its diff is empty, and left at `f87c529` it would move `main`'s pointer **backward** by four commits.

**Why it played out this way.** A gitlink-bump PR carries no content of its own — it only advances `main`'s pointer along `docs/internal`'s main. Any other PR that bumps the same pointer past yours delivers your content for you. That is not a conflict to resolve; it is the mechanism working. The required `pointer-check` enforces landing on the tip (R3), so a docs-only PR competing with actively-pushing sessions can be overtaken between its last re-bump and its merge. This one was overtaken twice in about forty minutes:

- `e69d87d66` → R3 failed, *"behind by 4 commits"*. Re-bumped to `f87c529` with `scripts/bump-docs-pointer.sh`.
- `f87c529` → overtaken again, this time by #2825, which carried ADR-153 to `main` itself.

**Nothing is lost.** The ADR is on `main`, at the same bytes this PR reviewed, having passed its cross-family gate here. SMI-6516 keeps its own status; this PR was the last deliverable its plan allocated, and that deliverable now exists.

Reopen if you disagree with the reading — the branch is untouched.

PR #2820, opened 2026-09-13
docs(plans): SMI-6497/6498/6499/6511 + ADR-156 -- register the reviewed plan revisions

wrsmith108:
## Business Summary

**What shipped:** Seven implementation plans and one architecture decision record, all reviewed and corrected. No code changed — these are gated plan-only deliverables. The review found real defects in every plan it looked at, including two in the fix for an earlier defect.

- Three plans that had never been reviewed came back with **3 BLOCKERs and 3 MAJORs**, including two plans that would have broken each other on landing.
- The fix for one of those introduced **two new fail-open paths**, caught by a follow-up round scoped to the fix itself.
- A plan previously reported as "ready to implement" turned out to contain a check that **would have failed CI on a clean tree**.

**Quality bar:** Every review ran on a different model family than wrote the plans (GPT-5.6-Sol, per ADR-128). Claims were executed rather than argued — git output shapes, shell exit semantics, regex predicates, normalizer behaviour, and workflow structure each run against a case table first. Several findings were *rejected* on measurement, including four of my own.

**Found along the way:** SMI-6583 (a silent red-fail path in the publish workflow, reproduced independently before filing), SMI-6568 and SMI-6569 (two guard defects, one with its mechanism proven deterministically), SMI-6576 (a release gate that can never decline). All filed separately rather than folded in.

**Net result:** All seven plans ready for implementation, in a fixed order. SMI-6493 closed on measured production evidence.

---

## Technical detail

One commit, one file: the `docs/internal` gitlink, `be0852e` → `5669823`. Forward-only, verified by hand, nothing dropped.

### Why one bump against the live tip

Two earlier PRs each pinned a *specific* `docs/internal` commit and were invalidated by sibling merges advancing main past them — once backward (R7), once diverged (R6). A pinned pointer is fragile by construction. This registers the live tip instead, and both earlier PRs are closed as superseded with that reasoning recorded.

### What the reviews found

**SMI-6512 and SMI-6513 would have broken each other.** SMI-6513 splits `publish.yml`'s block 4 at `/^SMOKE_DIR=/` and fails closed on zero matches. SMI-6512's fix rewrites that exact line. All three candidate anchors measured against the post-fix string return false, so SMI-6513 would hard-fail with no digest left to bump — and SMI-6512's stated remedy was to bump that digest. Resolved by an explicit marker comment located *before* normalization strips comments, locked by a test.

**SMI-6517's guard was inert.** `git diff --cached --raw` emits a leading colon on field 1, so `old_mode != 160000` was true for every gitlink modification and skipped all of them.

**SMI-6497's retirement step lost real protection.** Deleting SMI-6513's check is correct for the verify head, which extraction collapses — but the smoke tail survives almost verbatim, and the pinned digest was the only thing guarding it.

**ADR-156's central claim was false.** It argued a composite action has no test seam. A composite action can wrap a tested script. Restated as "shared CI logic must expose a runner-independent executable seam," with an honest comparison in which the composite wins two axes.

**The registry fix was itself fail-open, twice.** Stage-0 parse protection was dropped — an empty or comment-only `run:` passes silently — and "exact set" was exact only over a naming convention. Closed with a parse guard and three arms, with the residual named rather than hidden.

**SMI-6511's Wave 3 would have red-failed a clean main.** It keyed on steps containing `npm publish`; six steps match but only five jobs do, because an error-annotation step says "npm publish succeeded for …" in an `echo`. Now job-keyed.

### On that last one

I had reported SMI-6511 as gate-satisfied and unblocked. The SMI-6015 confirmation round that cleared it was correctly scoped to the collector reorder and correctly found it sound — but Wave 3 was never in scope. **A narrowly-scoped round clears its scope, not the document.** Corrected on the issue.

### Landing order

**SMI-6513 → SMI-6512 → SMI-6497.** Landing 6497 first makes SMI-6512's marker precondition return 0, which its own plan reads as "6513 hasn't landed" — a wrong diagnosis pointing at the wrong remedy. SMI-6498's Wave 3 is blocked on SMI-6513 only and is order-independent relative to SMI-6497.

### Not verified, stated rather than glossed

The P-7 state-flip scanner is absent at this worktree's strategy-submodule pin; a hand-rolled substitute was used and labelled. The GitHub Packages latency claim is cited in-repo but never independently confirmed — which is why SMI-6498's enterprise check is non-blocking and logs latency instead of asserting a threshold. None of the three plans is implemented, so every "fails closed" result is a property of the contract text, not of running code.

[skip-impl-check]


wrsmith108:
Closing. **The work is landed; only the pointer registration is outstanding, and it no longer needs this PR.**

All seven revised plans and ADR-156 are on `skillsmith-docs` and are an **ancestor of the current tip** — verified:

```
git -C docs/internal merge-base --is-ancestor 5669823 19fa9a4  -> 0
```

So **any** future pointer bump registers them automatically. Nothing can be lost by closing this, and no re-push is required to preserve it. Present at the tip: `smi-6497`, `smi-6498`, `smi-6499`, `smi-6511`, and `adr/156`.

**Why this PR cannot land as-is.** It pins `5669823`. The docs tip has since advanced to `19fa9a4` — three moves in about forty minutes — so `pointer-check` correctly reports R3 (behind). Re-bumping requires a `git push`, and pre-push in this worktree needs its own container. That container was SIGKILLed 36 hours ago (`ExitCode 137`, and per `docker inspect` **not** an OOM kill — I had been asserting OOM without checking) and its Docker network no longer exists, so a plain start fails and a `--force-recreate` would be needed. The host has been oscillating between load 1.8 and 8.0 with the Docker VM spiking to 543%, which is the wrong condition to add a container in.

Pushing with `SKILLSMITH_WORKTREE_PREPUSH_HARDFAIL_DISABLE=1` is not a workaround here: that variable falls back to **host execution**, and 44 container-aware tests then fail — `gitdir-writer-check`, `repair-worktrees-docker-guard`, `state-flip-scan` and others assert container-specific behaviour by design. `Test (root)` and `Test (root colocated)` are both green on main, so those failures are an artifact of the bypass, not a real signal. That is worth recording on its own: **the opt-out is a whole-pipeline execution-context switch, not a narrow container-gate skip.** Its first two uses looked free only because the diff took the docs-only path and nothing ran.

**What is left for whoever picks this up**: nothing specific to this branch. The next routine `docs/internal` bump carries it. If it is still behind later, `./scripts/bump-docs-pointer.sh` from a checkout with a working container is the one-step fix.

Related findings filed tonight: SMI-6568 (the pre-commit docs-only carve-out classifies already-committed work — and after a rebase the set is not merely empty but *wrong*, populated with other sessions' commits), SMI-6569 (the pointer-ancestry check reports a false R1 from any git hook, because `git -C` does not override an inherited `GIT_DIR`), SMI-6583, SMI-6576, and a recurrence datapoint on SMI-6383.

PR #2821, opened 2026-09-13
fix(mcp-server): SMI-6585 -- the install pre-flight reports its failure instead of swallowing it

wrsmith108:
## Business Summary

**What shipped:**

- When `install_skill`'s safety pre-flight cannot run, it now says so instead of continuing silently. The install still happens and is still checked — but you are told the check was skipped, why, and that the conflict action you asked for was not applied.

**What this is not:** it was never possible for a skill folder to be overwritten because of this. The installer runs the same target guard itself, before any download or disk write, so an unsafe target was always refused. What was missing was the report, not the protection.

**Quality bar:** the fix was reverted and the two new tests confirmed to fail (2 failed / 17 passed, naming exactly those tests), then restored byte-for-byte and confirmed back to 19 passed. Without that control, a passing test says nothing about whether the fix is what makes it pass. Full suites: 1,109 tests in the package and 1,973 in the src-colocated set, all green; typecheck and formatting clean.

**Found along the way:** nothing new. The 14 lint errors in the tree are all in the strategy submodule's test fixtures (SMI-6562, already filed) and none are in this diff; `audit:standards` exits 1 on the Check 69 crash that affects every worktree container (SMI-6575, already filed).

**Net result:** SMI-6585 is closed. The defect shipped in `@skillsmith/mcp-server@0.7.15` an hour after Wave A0 merged, and was found by a peer session that instrumented the catch rather than reading it — thirteen cross-model review rounds, six adversarial rounds and three pre-merge gate passes had all read this file without seeing it.

---

## Technical detail

Issue: SMI-6585. Introduced by SMI-6529 Wave A0 (`c97d2078f`, PR #2809).

### What was wrong

`packages/mcp-server/src/tools/install.ts` had a bare `catch {}` — *"Conflict check failed; proceed with normal install"* — that **predates** the guard it came to enclose. Wave A0 added `checkInstallTarget` **inside** that existing try:

```
+ const targetCheck = await checkInstallTarget({
```

A fail-closed guard reached through a catch that resumes is fail-open. The catch wrapped `loadManifest`, `manifestKeyFor`, `checkInstallTarget` and `checkForConflicts`; a throw from any of them was discarded with no error, no warning and no telemetry.

### Why it is Medium, not High

Core enforces the same guard again, unconditionally, at the top of both write paths — before any fetch or disk write:

| Path | Call site |
|---|---|
| `service.install()` | `skill-installation.service.ts:156` → `if (!targetCheck.ok) return buildInstallFailure(...)` |
| content install | `skill-installation.content.ts:331` — same shape |

`service.ts`'s only other empty catch (line 177) wraps `fetchFromGitHub` *after* the guard and returns a failure rather than proceeding; `content.ts` has none. The MCP tool delegates to `service.install()` immediately after the swallowing catch, so **no write path opened**.

### What the swallow actually cost

1. **The early exit — the pre-flight's entire purpose.** `checkForConflicts` runs `createSkillBackup` + `cleanupOldBackups`; on a swallowed throw those side effects ran for an install core was about to refuse. A0's own comment in that block says the pre-flight exists to refuse *before* they run.
2. **The caller's `conflictAction`**, silently not applied (`conflictCheck.earlyReturn` never reached).
3. **Diagnostics** — an EACCES `lstat` or a corrupt manifest (SMI-5909) became "guard skipped, install proceeds", logged nowhere.

### The fix

The failure is collected and surfaced through `tips`, matching the file's existing idiom for non-fatal problems (`alsoLinkFailures`, SMI-6529 round 7) rather than importing a helper from core:

```ts
const preflightProblems: string[] = []
...
} catch (err) {
  preflightProblems.push(`the install pre-flight (conflict check and target guard) could not be evaluated (...)`)
}
...
const nonFatalProblems = [...preflightProblems, ...alsoLinkFailures]
```

The message names the underlying error and the unapplied `conflictAction`, so it cannot degrade into a string a caller can do nothing with.

### The test change is part of the fix

The existing test asserted `toEqual(HAPPY_RESULT)` under the comment *"Conflict preflight swallows errors and continues with normal install"* — it encoded the defect as intended behaviour. It now asserts the failure is reported, and a second test pins that the report carries the real cause and the unapplied action.

**An assertion that the pre-flight RAN cannot distinguish "never ran" from "threw and was swallowed"** — both read as zero calls. That ambiguity is why the peer's investigation surfaced only `Number of calls: 0` with no reason, and it is a large part of why the defect survived every review round.

### Review-process note

A diff-scoped review structurally cannot see this class: the reviewer reads the added guard, while the catch two screens below renders as unchanged context — because it *is* unchanged. Its meaning inverted without its text changing. Recorded on SMI-6488, with a concrete checklist item: for every added call inside an existing block, read outward to the nearest enclosing `try`/`if`/early return and ask whether that construct still means what it meant before.

### DIFF
--- a/packages/mcp-server/CHANGELOG.md
+++ b/packages/mcp-server/CHANGELOG.md
@@ -4,6 +4,28 @@ All notable changes to `@skillsmith/mcp-server` are documented here.
 
 ## [Unreleased]
 
+- **Fix**: `install_skill` now reports it when its pre-flight safety check could not
+  run, instead of continuing silently. The pre-flight sat inside a bare `catch {}`
+  that predates it; SMI-6529 Wave A0 moved the `checkInstallTarget` guard inside that
+  catch, which made a fail-closed guard fail open for anything that threw there — a
+  manifest load, the target guard itself, or the conflict check.
+
+  **No skill folder was ever at risk from this.** The installer runs the same target
+  guard unconditionally before any download or disk write, so an unsafe target was
+  always refused. What was missing was the report, not the protection. The install's
+  outcome is deliberately unchanged; what is no longer invisible is that the
+  pre-flight did not complete — which also means its conflict backup was skipped and
+  a requested `conflictAction` may not have been fully applied. That now rides
+  `tips`, the same surface fan-out refusals already use, and names the underlying
+  error so a caller can tell "pre-flight passed" from "pre-flight could not be
+  evaluated" (SMI-6585).
+
+- **Fix**: namespace-gate warnings no longer replace the installer's own warnings in
+  the tool result; both sets are merged. Pre-existing since SMI-4588 and harmless
+  while only one producer ever returned warnings — the same shape as the catch
+  above, where a line's meaning inverts the day something else starts returning
+  data (SMI-6585, cross-model review).
+
 ## v0.7.15
 
 - **Fix**: SMI-6508 -- move MF-5 last; it was suppressing HIGH findings (#2808)
--- a/packages/mcp-server/src/tools/install.test.ts
+++ b/packages/mcp-server/src/tools/install.test.ts
@@ -20,9 +20,17 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 // Zod gate + delegation. `vi.hoisted` is required because `vi.mock` is
 // hoisted above regular `const` declarations and would otherwise reference
 // the stubs before they are initialised.
-const { mockInstall, mockEmitInstallEvent } = vi.hoisted(() => ({
+// SMI-6585 cross-model review: `checkInstallTarget` was reached through the
+// `...actual` passthrough, so no test could make the TARGET GUARD itself
+// throw — only the manifest load. That left three of the pre-flight's four
+// throw sources uncovered, and a regression that reported manifest failures
+// while silently losing guard failures would have passed. It is stubbed here,
+// defaulting to the same `{ ok: true }` the real guard returns for these
+// fixtures, so existing tests are unaffected and the throw path is reachable.
+const { mockInstall, mockEmitInstallEvent, mockCheckInstallTarget } = vi.hoisted(() => ({
   mockInstall: vi.fn(),
   mockEmitInstallEvent: vi.fn(),
+  mockCheckInstallTarget: vi.fn(),
 }))
 
 vi.mock('@skillsmith/core', async (importActual) => {
@@ -34,6 +42,7 @@ vi.mock('@skillsmith/core', async (importActual) => {
     ...actual,
     SkillInstallationService: MockSkillInstallationService,
     emitInstallEvent: mockEmitInstallEvent,
+    checkInstallTarget: mockCheckInstallTarget,
   }
 })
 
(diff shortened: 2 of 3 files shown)


wrsmith108:
## Cross-model pre-merge gate: GPT-5.6-Sol via NEEDLE (ADR-128)

Dispatched against a clean read-only workspace containing the full diff, both changed files, and excerpts of the two core guard call sites. Outcome `success`, ~69s, 138k input / 3.3k output tokens.

**The central claim was independently confirmed.** The reviewer verified that the MCP wrapper's only write-capable delegation is `service.install`, that both the registry and content call sites place `checkInstallTarget` before their first fetch/scan/write, that `preflightProblems` is unconditionally initialised with no early return able to discard it, and that existing `result.tips` are preserved. Its words: *"Medium rather than High is therefore justified."* It also swept the file for sibling silent catches and found none.

Five Low findings. **All five are addressed in this PR — none deferred.**

| # | Finding | Disposition |
|---|---|---|
| 1 | `String(err)` can itself throw (`Object.create(null)`, hostile `toString`), converting a swallowed failure into an escaped exception | **Fixed** — non-throwing formatter with fallback, plus a test that rejects with a throwing-`toString` object |
| 2 | Raw error text can disclose local paths | **Partly accepted** — see below |
| 3 | "conflictAction was not applied" claims more than the code can know | **Fixed** — now "may not have been fully applied"; `checkForConflicts` can throw *after* writing a backup |
| 4 | Both tests exercised only `loadManifest`, leaving 3 of 4 throw sources uncovered | **Fixed** — `checkInstallTarget` is now stubbed (it was unreachable behind `...actual`), with separate tests for the target guard, the conflict check, and a non-`Error` value |
| 5 | Namespace warnings *replace* installer warnings | **Fixed** — merged, same as the `tips` path |

### On finding 2, where I partly disagree

The `tips` surface returns to the MCP caller — the local operator who just ran the install on their own machine. A filesystem path in that message is diagnostic value, not disclosure to a third party, and redacting it would defeat the change's entire purpose: the defect was that failures were invisible.

What I did accept is the unbounded part. An arbitrary throw site can produce an arbitrarily long string, and passing that through unexamined is careless regardless of its content, so the message is now capped at 300 characters. I did not add redaction, and I'd rather record the disagreement than silently half-comply.

### On finding 5, worth noting

The reviewer spotted that the warnings-replace bug is *the same shape as the defect this PR fixes* — a line whose meaning inverts the day a producer elsewhere starts returning data it previously never did. That is pre-existing code from SMI-4588, not mine, but it is in the file I was changing and it is the same class, so it is fixed here rather than filed.

### Verification after the review fixes

- Negative control extended across every throw source — with the catch reverted to its silent form, all reporting tests fail; restored byte-for-byte, green again.
- 22 tests in the file (was 18 before this PR), full mcp-server package suite 1109 passed, typecheck exit 0, format clean.

### One process note

Codex auto-discovered a `commit` skill in its sandbox and spent part of its run intending to commit the review — despite the prompt stating analysis-only. It could not (read-only sandbox, workspace is not a git repo). That is the workspace-auto-discovery hazard CLAUDE.md warns about for NEEDLE dispatch, observed live.

PR #2823, opened 2026-09-13
fix(ci): SMI-6580/6569 -- pointer-guard integrity: a failed alert must fail the job, and the local check must read the right repo

wrsmith108:
## Business Summary

**What shipped:**

- `pointer-autorepair` now fails its job when it cannot deliver its alert, and names the cause in the log. It previously emitted a warning and reported success.
- The GitHub label that alert depends on now exists. It never had, so every alert this job has attempted since it shipped failed with `could not add label: 'submodule-pointer-regression' not found`.
- The pre-push submodule-pointer check now reports the correct verdict when run from a worktree. It previously reported `R1: was never pushed` for commits that are pushed and reachable, and because that check is a precondition, the false finding also suppressed the true one.

**Quality bar:** Plan-review (ADR-109 Stage 2) returned BLOCKED with 7 blockers, which cut the plan from four waves to two — two directions turned out to duplicate existing plans of record. Governance review then found one MAJOR issue in the first implementation, fixed here. Every behavioural claim in the diff was executed rather than reasoned about, and both regression tests were verified three ways: passing on the fix, **failing** on the reverted code, passing again on restore.

**Found along the way:**

- **SMI-6587 (Urgent)** — the required PR gate `pointer-check` fails *open*: when the `docs/internal` fetch fails, a real ancestry violation becomes a green required check. Demonstrated on identical trees (exit 1 with the submodule present, exit 0 without). It changes a required check's accept predicate, so it needs its own SPARC + plan-review under ADR-109 and is filed rather than fixed here.
- **SMI-6586** — two monitors name a GitHub label that nothing creates. This PR creates only the one it needs; the class, and the `audit:standards` check that would stop it recurring, belong to that issue.
- **SMI-6582** re-routed to SMI-6502 and **SMI-6581** to ADR-154, both of which already own those questions. The measured reasoning is preserved in the plan rather than deleted with the waves.

**Net result:** Both waves shipped. Three follow-ups tracked, one Urgent, none blocking this PR.

---

### Wave 1 — `6686c3ab5`, SMI-6580

`open_or_skip_issue()` in `scripts/ci/submodule-pointer-autorepair.sh` had five paths that returned success on a failed alert:

| Line | Before | After |
|---|---|---|
| `:92-95` | `gh issue create … \|\| echo "::warning::"` | capture stderr, `::error::` with the real cause, return non-zero |
| `:77-79` | dedupe `gh issue list … \|\| echo ""` | query failure distinguished from "no existing issue" |
| `:61-65` | non-zero inner exit with no `FAIL` line → "nothing actionable" | exit 2 (broken invocation) distinguished from exit 1 (content verdict) |
| `:131`, `:149` | every path `exit 0` | status propagated |
| — | label assumed to exist | idempotent `gh label create`, plus a registry entry in `setup-github-labels.sh` |

Ordering is load-bearing: failing the step *without* creating the label first would leave the job permanently red while still opening no issue.

The script was restructured into `main()` behind a `BASH_SOURCE` guard so `open_or_skip_issue` can be unit-tested with a faked `gh` on `PATH`. It had zero test coverage; it now has 7 tests.

**One defect was introduced by this change and caught before commit.** The dedupe fix first captured `$(gh issue list … 2>&1)`, merging stderr into the value. `gh` writes advisory notices to stderr on *successful* calls, so a run where no issue existed came back non-empty and read as "already reported" — reintroducing the exact failure being removed. stderr now goes to a file.

### Wave 2 — `79b8e159b` + `1ccd0fb95`, SMI-6569

`git -C <dir>` does not override an absolute `GIT_DIR`: `-C` changes the working directory, but `GIT_DIR` still wins over repo discovery. Git exports an absolute `GIT_DIR` into hooks on a push from a **linked worktree**, which is this repo's default workspace.

```
GIT_DIR unset             -> git -C docs/internal cat-file -e <sha>^{commit} -> exit 0
GIT_DIR=<absolute>        -> same command                                    -> exit 128
GIT_DIR=".git" (RELATIVE) -> same command                                    -> exit 0
```

The effect is absolute-only; a relative `GIT_DIR` re-resolves against `-C`'s cwd. Scope is the whole helpers file, not the one line the symptom surfaced on — under contamination the fetch, the `origin/<branch>` rev-parse, and every `merge-base`/`rev-list`/`branch` comparison were also reading the outer repo. R1 fired first and masked them.

All 9 submodule-directed calls in `check-submodule-pointer.helpers.sh` route through a new `git_sub()`; the 2 in `submodule-pointer-autorepair.sh` route through a local `git_mount()`. Outer-repo calls are deliberately **not** wrapped — verified that `rev-parse --show-toplevel` and `ls-tree HEAD` return identical results with and without `GIT_DIR`, and that a bogus discovery var makes the script exit 2 with `not inside a git repository` rather than return a wrong verdict.

`1ccd0fb95` widens both wrappers from 2 vars to the 10 in `GIT_DISCOVERY_VARS` (`scripts/tests/_lib/git-fixture-env.ts`, SMI-4693), after governance confirmed `GIT_COMMON_DIR` and `GIT_OBJECT_DIRECTORY` redirect straight through a `GIT_DIR`-only unset. That list's last two entries, `GIT_CONFIG` and `XDG_CONFIG_HOME`, are deliberately not unset: they route config rather than discovery, and these wrappers run an authenticated `fetch` whose credentials come from `git config --global`. Measured that unsetting them would have been safe either way, so this is a scoping choice rather than a workaround.

### Testing

Both test files run on the host (`npx vitest run <file>`) — they spawn `bash` and need no container.

- `scripts/tests/submodule-pointer-autorepair.test.ts` — new, 7 tests.
- `scripts/tests/check-submodule-pointer.test.ts` — 23 tests, one new, asserting the invariant across `GIT_DIR`, `GIT_COMMON_DIR` and `GIT_OBJECT_DIRECTORY`.

The Wave 2 test asserts **stdout and exit-status equality** rather than a symptom name. Which symptom appears without the fix depends on whether the outer repo has an `origin` remote: production produces the fabricated `R1`, the fixture produces `R-FETCH` because the misdirected fetch fails outright. Asserting either name alone would let the test pass against the bug in one of the two environments.

### CI caveats

- No `workflow_dispatch` was added. `pointer-autorepair` fires only on push to `main`, so the plan's original acceptance test ("induce a failure on a scratch branch") was not executable; the binding acceptance is the unit tests, which do not need it.
- `shellcheck` is clean on all three shell files, but no CI job covers them — `ci.yml`'s shellcheck is scoped to `scripts/smoke-prod/*`.
- `SC2148` on `check-submodule-pointer.helpers.sh` is pre-existing and expected (sourced-only file, no shebang by design); verified against the committed baseline.
- The two `docs/internal` pointer commits carry the plan-review-amended implementation plan. The second is a re-bump: `docs/internal` main advanced mid-review, which is the SMI-6582 treadmill this PR's own plan documents, owned by SMI-6502.

### Plan

`docs/internal/implementation/smi-6580-6582-pointer-guard-integrity.md` (amended `55d0778`).

### DIFF
--- a/scripts/ci/check-submodule-pointer.helpers.sh
+++ b/scripts/ci/check-submodule-pointer.helpers.sh
@@ -75,6 +75,48 @@ repo_slug() {
     printf '%s:%s' "$_CSP_SLUG" "$_CSP_BRANCH"
 }
 
+# git_sub <submodule-dir> <git-args...> — run git against a SUBMODULE working
+# directory, immune to an inherited GIT_DIR (SMI-6569).
+#
+# `git -C <dir>` does NOT override GIT_DIR: -C changes the working directory,
+# but an ABSOLUTE GIT_DIR still wins over repo discovery, so the command runs
+# against the OUTER repo while appearing to target the submodule. Git exports
+# an absolute GIT_DIR into hooks on a push FROM A LINKED WORKTREE — this
+# repo's default workspace — so `.husky/pre-push`'s invocation hit this on
+# every run. Measured on git 2.50.0 against docs/internal:
+#
+#   GIT_DIR unset             -> git -C docs/internal cat-file -e <sha>^{commit} -> exit 0
+#   GIT_DIR=<absolute>        -> same command                                    -> exit 128
+#   GIT_DIR=".git" (RELATIVE) -> same command                                    -> exit 0
+#
+# A relative GIT_DIR re-resolves against -C's new cwd and is harmless; only an
+# absolute one redirects. End-to-end on one identical tree, the symptom was a
+# fabricated R1 ("was never pushed") for a commit that is pushed and reachable,
+# which ALSO suppressed the true verdict (R3) because R1 is a Layer-1
+# precondition that short-circuits Layer 2 entirely.
+#
+# Every git call in this file that targets the submodule must go through this
+# wrapper, not just the R1 existence check: under a contaminated GIT_DIR the
+# fetch, the origin/<branch> rev-parse, and every merge-base/rev-list/branch
+# comparison were all reading the OUTER repo too. R1 simply failed first and
+# masked the rest.
+#
+# Calls targeting the OUTER repo (`git -C "$_CSP_ROOT" ls-tree`, and
+# check-submodule-pointer.sh's own rev-parse/diff) deliberately do NOT use
+# this wrapper: they want the outer repo, and an inherited worktree GIT_DIR
+# already names it. Verified — `rev-parse --show-toplevel` and `ls-tree HEAD`
+# return identical results with and without GIT_DIR set.
+#
+# GIT_WORK_TREE is unset alongside GIT_DIR as defence in depth. Measured, only
+# GIT_DIR is actually exported into hooks, and unsetting GIT_DIR alone is
+# sufficient; unsetting both costs nothing and removes the remaining way a
+# caller's environment could redirect these reads.
+git_sub() {
+    _CSP_GIT_SUB_DIR="$1"
+    shift
+    env -u GIT_DIR -u GIT_WORK_TREE git -C "$_CSP_GIT_SUB_DIR" "$@"
+}
+
 # print_result — unified output line. severity: PASS|PASS-WARN|SKIP|FAIL.
 # is_blocking (0/1) downgrades a FAIL's displayed severity to WARN (mount
 # not in BLOCKING_MOUNTS) without changing the caller's exit-code decision,
@@ -125,7 +167,7 @@ evaluate_mount() {
     # Not-initialized check: test for a LITERAL .git entry directly under the
     # mount directory (file or dir — a submodule's own .git is a file
     # pointing at ../../.git/modules/<path> since git 1.7.8+). Deliberately
-    # NOT `git -C "$_CSP_DIR" rev-parse --git-dir`: for an uninitialized
+    # NOT `git_sub "$_CSP_DIR" rev-parse --git-dir`: for an uninitialized
     # mount (an empty directory with no .git of its own), git's repo
     # discovery walks UP the directory tree and silently finds the PARENT
     # repo's own .git instead of failing — that would make this check
@@ -144,7 +186,7 @@ evaluate_mount() {
     _CSP_FETCH_OK=0
     _CSP_I=0
     while [ "$_CSP_I" -lt "$_CSP_ATTEMPTS" ]; do
-        if git -C "$_CSP_DIR" fetch origin --prune --quiet 2>/dev/null; then
+        if git_sub "$_CSP_DIR" fetch origin --prune --quiet 2>/dev/null; then
             _CSP_FETCH_OK=1
             break
         fi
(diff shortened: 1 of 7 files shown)


wrsmith108:
## Pre-merge gate: 15 checks, cross-family (ADR-128)

Dispatched to **GPT-5.6-Sol via NEEDLE**, not a same-family Claude subagent, because this PR was Claude-authored.

| Check | Result | Note |
|---|---|---|
| PR-01 Migration ordering | SKIP | No SQL migrations in the diff |
| PR-02 RLS → policy | SKIP | No SQL |
| PR-03 DROP TABLE FK | SKIP | No SQL |
| PR-04 `search_path` | SKIP | No SQL functions |
| PR-05 `schema_version` seq | SKIP | No migrations |
| PR-06 Lock profile | SKIP | No DDL |
| **PR-07 Silent catch / write paths** | **FAIL → FIXED** | Four remaining fail-open paths found; all fixed in `28c7f170f` |
| PR-08 Upsert callers | SKIP | No column changes |
| PR-09 Surface registration | SKIP | No edge function; label registered on both planned surfaces |
| PR-10 Export continuity | PASS | `BASH_SOURCE` guard preserves direct-execution behaviour; required globals set before every dispatch |
| PR-11 Enum consumers | SKIP | No enum changes |
| PR-12 Plan vs diff | PASS | Waves 1–2 represented; withdrawn Waves 3–4 did not leak in |
| PR-13 SQL comment accuracy | SKIP | No SQL |
| PR-14 `audit:standards` | PASS | 89 passed / 7 warnings / 0 failed — unchanged baseline, no warning references a changed file |
| PR-15 State-flip backstop | PASS | Label registration is additive; no default or guard state flipped |

## Findings, all resolved in `28c7f170f`

Two were CONFIRMED with executed counterexamples, and both were things two Claude-side review passes had missed.

**F-1 (High) — the outer-repo calls were not safe.** The previous commit wrapped only submodule-directed calls, on the reasoning that an inherited worktree `GIT_DIR` already names the outer repo. That holds only for the value a hook exports. Pointing `GIT_DIR` at a *different valid repository*:

```
clean     ls-tree HEAD -- docs/internal  →  160000 commit f87c529… docs/internal
poisoned  ls-tree HEAD -- docs/internal  →  (empty), exit 0
```

An empty `S` is read by `evaluate_mount` as "no gitlink entry … nothing to check" and returned as **PASS** — a wrong answer delivered as a pass. In `submodule-pointer-autorepair.sh` the same class reaches `add`, `commit` and `push`.

Fixed by sanitizing **once at each entry point**, before the repo root is resolved, rather than enumerating call sites.

**F-3 (Medium) — the variable list was the wrong contract.** `GIT_SHALLOW_FILE` is not a discovery variable and redirects nothing; it rewrites *ancestry*, which is what R3–R7 are decided on:

```
merge-base --is-ancestor OLD TIP   0 → 1
rev-list --count OLD..TIP          3 → 2
```

`GIT_GRAFT_FILE` rewrites parentage the same way. Both added; the contract is now "verdict-affecting repository environment", not "discovery".

**F-2 (High)** — the "non-zero exit, no FAIL line" branch was a denylist that special-cased exit 2 and returned 0 for everything else (126, 127, signal death). Inverted to an allowlist: only exit 1 is a content verdict. `grep`'s own status is captured instead of `| head -1 || true`.

**F-4 (Medium)** — dedupe matched the prose `commit <sha>` in any body with `--state all`, so an old closed write-up mentioning that SHA could suppress a real alert. Now keyed on an invisible marker and constrained by label.

**F-5 (Low)** — `gh label create || true` made a failed label write indistinguishable from success; both `mktemp` calls unchecked; `main()`'s temp file leaked. All fixed.

**F-6 (Low)** — two copies of the variable list synchronized by a comment. Both now delegate to `scripts/ci/git-env-sanitize.sh`.

## A test in this PR was decorative and its own control caught it

The first `GIT_SHALLOW_FILE` case pointed at `/dev/null`, which flips `--is-shallow-repository` but lists no boundary and truncates nothing — it **passed with the fix removed**. The boundary has to be the *advancing* commit, since shallow cuts a commit's parents, so detaching `T` leaves `T → newTip` intact.

Control now reads **fixed=0 / reduced=1 / restored=0**, and the failure shows the real corruption: **R6 "diverged" reported where the truth is R3 "stale"**.

## Filed, not fixed here

- **SMI-6587 (Urgent)** — `pointer-check` fails *open*: a failed `docs/internal` fetch turns a real ancestry violation into a green required check. ADR-109 applies.
- **SMI-6586** — two monitors name a GitHub label nothing creates. This PR creates only the one it needs.
- **SMI-6595** — shellcheck coverage is a per-file allowlist, so these four scripts are linted by nothing in CI. They are clean; nothing enforces it.
- **SMI-6506** closed as a duplicate of SMI-6569 — it had hypothesized this exact cause and marked it unverified.

PR #2886, opened 2026-09-17
chore(docs): bump docs/internal pointer to 58e244a — SMI-6714 round 3 + three-round retro

wrsmith108:
## Business Summary

**What shipped:** One pointer line, carrying six documents that were already written and already safe — and closing a gap where this repository's `main` had drifted six commits behind the internal docs it references.

The substance is a plan for the P-7 state-flip scanner, the tool a PR review runs to find assertions a state flip has silently invalidated. **It went through three review rounds today and each one found a defect inside the previous round's fix.** The third found one inside the artifact added to *prove* the second's fix worked — a worked example whose own formatting broke in exactly the way the fix existed to prevent, while its prose claimed the example was verbatim from real output. It wasn't: one record truncated, one with its indentation collapsed, one with the search term itself deleted, so as printed it could never have come from that search at all.

Also carried: a retro recording what transferred, because three rounds of this is worth more than one plan. Ten items, each with the measurement behind it rather than stated as advice. The two that generalise furthest:

- **A check's current value is not its discriminating power.** One criterion was called "the strongest" on the strength of its score. Renaming two variables — changing no logic — moved that score from 9 to 3. Rewriting the same flawed code in different idioms scored a perfect 0 while producing byte-identical broken output. It would also have scored 0 for a design the plan *rejected on the strength of that very check*. Nobody had ever tested whether the number meant anything.
- **Absence and presence are different questions.** Replacing a check that asks "is the bad thing gone?" with one that asks "is the good thing present?" closes nothing — both can be true at once. Proven by building exactly that: a program passing every gate while keeping the flaw intact.

**Quality bar:** every applied finding was re-measured independently before being written down. **Two of my own measurements were invalid and are recorded rather than quietly redone** — one mutation changed nothing and I read the unchanged result as evidence, which is the same defect the work is about. Three renderers were measured, not one, and a third disagreed sharply with the other two.

**Honest about what this is not:** all three review rounds ran on the same model family that wrote the plan. That is not the independent review the rules require, and the plan says so wherever its review status appears. **SMI-6717** carries the cross-family pass after 2026-09-19, with the four questions the rounds could not settle about themselves.

**Net result:** six documents registered, `main` no longer drifting behind, and a plan that is markedly stronger and still not approved for implementation. Nothing executable changes.

---

## Technical detail

Registered range `51b5a9f..58e244a` — six commits, **every path `.md`**, verified with `git -C docs/internal diff --name-only`.

| | |
|---|---|
| mine | SMI-6714 plan, 648 lines after round 3; retro, 117 lines; `code_review/index.md`; root folder index |
| peer's | three SMI-6676 closeout-plan commits, one SMI-6724 plan |

Originally targeted `abd0d19`. A peer noted `docs/internal` main had advanced past it and that one bump to the tip delivers both sets. Retargeted after verifying its claims rather than accepting them: `abd0d19` is an ancestor of `58e244a`, `58e244a` is that submodule's actual `origin/main`, `51b5a9f → 58e244a` is a forward bump, and both my documents are present at the new target. My original target would have merged already-superseded, leaving `main` three behind.

### Round 3's two BLOCKING findings, measured

```
specimen, as round 2 wrote it   ->  code, paragraph, code   <table>:0
  (outer fence 3 backticks around an inner 6; CommonMark closes on >=)
specimen, corrected (outer 7)   ->  heading, paragraph, code, table
  block a reviewer pastes       ->  <pre>:1  <table>:1   0 of 3 records lost

criterion 1 grep, pristine ......... 9
  rename two walk variables ........ 3    bash -n clean
  rewrite parse in cut -d: idioms .. 0    byte-identical corrupt output
```

Three renderers on the raw output: `marked` 18/44 unrecoverable, `markdown-it` 18/44, **`pandoc -f gfm` 44/44** — its gfm reader reads `$…$` as TeX math and loses every record. All three: 0/44 fenced. `cmark-gfm`, Linear and GitHub comment bodies remain unmeasured; that is SMI-6717's first deliverable.

### Found along the way, filed not narrated

- **SMI-6501** — a shared container reporting `health=healthy` for 46 hours with zero restarts while eight `node_modules` mounts were detached; plus a peer's container detached on nine paths **97 minutes after a recreate**, which means the documented remedy buys an unmeasured window rather than a fix.
- **SMI-6720** — three review-rubric checks no round currently makes.
- **SMI-6717** — the cross-family pass, gated on Codex availability.
- **SMI-6527** — a second live reproduction of `bump-docs-pointer.sh` reporting "nothing to do" while leaving the pointer stale. I filed a duplicate of this before searching, and cancelled it.

### Gate carve-outs, stated rather than assumed

**`pr-reviewer` 16-check gate — SKIPPED** under the SMI-6701 carve-out, which turns on the registered *range*: no `.sh`, `.ts`, `.mjs`, `.sql` or workflow YAML. Executed, not inferred. **`/governance` — SKIPPED**, recorded in the commit; one gitlink line, no code to review. **`verify-implementation`** correctly fails a PR referencing an SMI with no source changes.

[skip-impl-check]


wrsmith108:
Closed deliberately, not abandoned — recording why so the next reader does not reconstruct it.

**Everything this PR carried is already durable on `docs/internal` main at `554670b`:** the SMI-6714 guard-surface plan (648 lines, three review rounds applied) and the three-round retro at `code_review/2026-09-17-smi-6714-three-round-review-retro.md` (117 lines), plus both index updates. A `docs/internal` pointer bump is disposable by design — the content lands in that submodule first and the pointer is bookkeeping, so the next session's bump delivers it.

**Why it closed:** the head ref was deleted while the PR was still open. The merge had been refused (`mergeStateStatus=BEHIND` — one commit behind under strict branch protection) and I ran the branch-delete unconditionally in the same command block instead of gating it on the merge result. GitHub auto-closes a PR whose head ref disappears. Nothing was lost; the commit survives locally.

**Why it was not relanded:** the pre-push gate blocked three further attempts on conditions unrelated to the diff, which is one gitlink line:

- `scripts/tests/git-crypt-remediation-strings.test.ts` T11 times out non-deterministically at its own 60s cap — measured 180s / 127s / 149s across three consecutive runs on an identical tree, two failures and one pass. Filed as **SMI-6726** with the measurements and a `git grep`-based fix proposal.
- `skillsmith-dev-1` re-detached eight `node_modules` mounts **111 minutes** after I recreated it, while reporting `health=healthy` with zero restarts. A peer measured 97 minutes on a different container. Two independent samples on **SMI-6501**: the documented recreate remedy has a roughly 100-minute half-life and nothing signals when it lapses.

Continuing would have meant recreating the shared container again — dropping both MCP servers for every session on this machine — to land a commit that is disposable. The owner chose to stop, which is the right trade.

**For whoever lands the bump:** the commit is ready and ancestry-validated. Target `554670b`, range `51b5a9f..554670b`, every path `.md`, forward ancestry confirmed. It will need a rebase onto current `main` and a re-push; expect T11 to need a retry or two until SMI-6726 is fixed.

